PR #37, opened 2026-01-11
Add Version, BasePath and VersionHandshakeStatus to IMcpPlugin public…

mainVar:
This PR enhances the IMcpPlugin public API by adding three new properties that provide better visibility into the plugin's runtime state and configuration:

Version - Access to API and plugin version information
BasePath - Actual server execution path (independent of launch location)
VersionHandshakeStatus - Real-time status of version compatibility handshake with the server
Also added a parameterless constructor to McpPluginBuilder for backward compatibility.

Changes
Core Interface Updates
IMcpPlugin.cs - Added three new public properties:
Common.Version Version { get; } - Returns version info (Api, Plugin, Environment)
string BasePath { get; } - Returns the actual execution directory
VersionHandshakeResponse? VersionHandshakeStatus { get; } - Returns the last handshake result
Hub & Connector Updates
IRemoteMcpManagerHub.cs - Added VersionHandshakeStatus { get; } property
BaseHubConnector.cs - Implemented handshake response storage and exposure:
Stores the last VersionHandshakeResponse in _lastHandshakeResponse
Exposes it via public VersionHandshakeStatus property
Initializes with default values when handshake fails
Implementation Details
McpPlugin.cs - Implemented new properties:
Version - Injected via DI from McpPluginBuilder
BasePath - Uses AppDomain.CurrentDomain.BaseDirectory
VersionHandshakeStatus - Delegates to RemoteMcpManagerHub
McpPluginBuilder.cs - Added parameterless constructor that creates a default Version() instance for backward compatibility

All existing tests pass (139/139) without modifications, confirming backward compatibility.

### DIFF
--- a/McpPlugin/src/McpPlugin/Network/Client/BaseHubConnector.cs
+++ b/McpPlugin/src/McpPlugin/Network/Client/BaseHubConnector.cs
@@ -29,7 +29,7 @@ namespace com.IvanMurzak.McpPlugin
         protected readonly CancellationTokenSource _cancellationTokenSource = new();
 
         private readonly ThreadSafeBool _isDisposed = new(false);
-        private volatile VersionHandshakeResponse lastHandshakeResponse = new();
+        private volatile VersionHandshakeResponse? lastHandshakeResponse = null;
 
         /// <summary>
         /// Disposable for subscription on the HubConnection changes.
@@ -199,13 +199,7 @@ namespace com.IvanMurzak.McpPlugin
             if (cancellationToken.IsCancellationRequested)
                 return;
 
-            lastHandshakeResponse = handshakeResponse ?? new VersionHandshakeResponse
-            {
-                ApiVersion = "Unknown",
-                ServerVersion = "Unknown",
-                Compatible = false,
-                Message = "Handshake failed: No response"
-            };
+            lastHandshakeResponse = handshakeResponse;
 
             if (handshakeResponse != null && !handshakeResponse.Compatible)
             {
--- a/McpPlugin.Tests/Mcp/McpBuilderTests.cs
+++ b/McpPlugin.Tests/Mcp/McpBuilderTests.cs
@@ -232,7 +232,7 @@ namespace com.IvanMurzak.McpPlugin.Tests.Mcp
         }
 
         [Fact]
-        public void Build_WhenNotConnected_ShouldHaveDefaultVersionHandshakeStatus()
+        public void Build_WhenNotConnected_ShouldHaveNullVersionHandshakeStatus()
         {
             // Arrange
             var reflector = new Reflector();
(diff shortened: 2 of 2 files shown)


McpPlugin/src/McpPlugin/Network/Client/BaseHubConnector.cs
```diff
@@ -29,6 +29,7 @@ public abstract class BaseHubConnector : IConnectServerHub, IDisposable
         protected readonly CancellationTokenSource _cancellationTokenSource = new();
 
         private readonly ThreadSafeBool _isDisposed = new(false);
+        private VersionHandshakeResponse _lastHandshakeResponse = new();
```
IvanMurzak: Need to use `_` only for `readonly` properties. This one should not have it.

McpPlugin/src/McpPlugin/Network/Client/BaseHubConnector.cs
```diff
@@ -194,6 +196,14 @@ private async void OnHubConnectionChanged(HubConnection? hubConnection)
                 },
                 cancellationToken: cancellationToken);
 
+            _lastHandshakeResponse = handshakeResponse ?? new VersionHandshakeResponse
```
IvanMurzak: It should be done after checking `cancellationToken.IsCancellationRequest`. Because async operation may be interrupted with incorrect result by the cancellation token, need to respect that scenario. Move this new block of code after the if-check

McpPlugin/src/McpPlugin/Network/Client/BaseHubConnector.cs
```diff
@@ -29,6 +29,7 @@ public abstract class BaseHubConnector : IConnectServerHub, IDisposable
         protected readonly CancellationTokenSource _cancellationTokenSource = new();
 
         private readonly ThreadSafeBool _isDisposed = new(false);
+        private volatile VersionHandshakeResponse lastHandshakeResponse = new();
```
IvanMurzak: - [ ] Until we received the connection handshake from the server we don't know VersionHandshakeReponse values. It should be `null` instead of default values.
- [ ] The property should be nullable as well (add `?`).

McpPlugin/src/McpPlugin/Network/Client/BaseHubConnector.cs
```diff
@@ -197,6 +199,14 @@ private async void OnHubConnectionChanged(HubConnection? hubConnection)
...
                 return;
 
+            lastHandshakeResponse = handshakeResponse ?? new VersionHandshakeResponse
+            {
+                ApiVersion = "Unknown",
+                ServerVersion = "Unknown",
+                Compatible = false,
+                Message = "Handshake failed: No response"
+            };
+
```
IvanMurzak: Lets leave it null, it if is null. Don't need to create a default value.

IvanMurzak (approved):
LGTM!

PR #202, opened 2026-08-15
feat(agent-config): machine credential store v2 — families schema, atomic writes, unreadable state

IvanMurzak:
Taskflow **unified-machine-auth**, task **b1-cs-store-v2** (security-critical). Implements the v2 store contract from design `04-credential-store-and-refresh.md` §1.

## What changed

- **v2 schema**: `families.{agent,plugin,legacy}` with per-family `clientId` + `scope` (`MachineCredentialFamilies` / `MachineCredentialFamily`).
- **v1 compat mirror**: every v2 write re-stamps the top-level `accessToken`/`refreshToken`/`expiresAt` from `families.plugin` (falling back to `families.legacy`; cleared for agent-only documents) so pre-v2 readers keep working during the transition window.
- **v1 read-compat**: a `{version:1}` document is interpreted as `families.legacy` on read; the first write upgrades it to v2 (+mirror). Versions newer than 2 pass through unchanged (no downgrade stamping).
- **Unknown-field preservation**: `[JsonExtensionData]` at the top level, the families container, and inside each family — fixes the shipped drop-on-read. Serializer options untouched (camelCase, null-omitting, indented); tests compare parsed values, never bytes.
- **Atomic writes** (replacing `File.WriteAllBytes`): temp sibling **in the same directory** → fsync → `File.Replace` (preserves the destination DACL) with a move fallback for the create case; **retry policy is part of the contract** — `RenameRetryAttempts = 5` (≥4), `RenameRetryDelayMs = 250` (≥250) on EBUSY/EPERM/EACCES-alike holders. Advisory parent-directory fsync on POSIX. The temp file is created `0600` before any secret byte lands.
- **Structured "store unreadable" state**: new `TryRead()` → `Missing` / `Ok` / `Unreadable`. DPAPI unprotect failure or a corrupted document never crashes, never deletes, never overwrites: `Rotate()` refuses on an unreadable store; an explicit re-auth `Write()` replaces it; `Read()` remains the lossy compat surface (null, never a throw). `Exists` documented as never being a signed-in signal.
- **`Rotate()` is family-aware**: rotates `families.plugin` (falling back to `legacy`, creating it on an empty store); the mirror follows on write.
- **DPAPI/POSIX unchanged**: CurrentUser, null entropy, `CRYPTPROTECT_UI_FORBIDDEN`; 0600/0700 re-asserted on the replace path.
- **CI**: the `test` job now runs on `ubuntu-latest` / `windows-latest` / `macos-latest` (the store's behaviour is OS-specific: DPAPI + rename/handle timing on Windows, permission asserts on POSIX). The docker-based result-publish step stays Linux-only; each leg uploads its own `.trx` artifact.

## Out of scope (per task boundaries)

- `ITokenRefresher`/refresh flow (task b3), lock protocol (task b2), NuGet release (task b4).

## Mandated plants (design 04 §5) — all RED-verified by mutation, then restored

| Plant | Mutation applied | Result |
|---|---|---|
| Atomicity | atomic write → direct `File.WriteAllBytes` | RED: concurrent readers observed truncation (5× "empty credential file") |
| Rename retry | retry filter disabled (single attempt) | RED: immediate sharing-violation `IOException` under a held destination handle |
| DPAPI corrupt fixture | `CryptographicException` catch removed | RED: 3 unreadable-state tests crash with `CryptographicException` |
| Unknown fields | `[JsonExtensionData]` removed at each of the 3 levels (3 runs) | RED each time: `KeyNotFoundException` on the preserved field |

Full suite after restore: **3032 passed, 0 failed** locally on Windows (net8.0 + net9.0; DPAPI and handle-holder legs exercised for real).

### DIFF
--- a/.github/workflows/test-pull-request.yml
+++ b/.github/workflows/test-pull-request.yml
@@ -7,7 +7,14 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    # Cross-platform matrix (design 04 §5): the machine credential store's at-rest protection and
+    # atomic-write behaviour are OS-specific — DPAPI + rename/handle timing on Windows, 0600/0700
+    # permission asserts on the POSIX legs — so the suite runs on all three.
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     permissions:
       checks: write
       pull-requests: write
@@ -36,8 +43,10 @@ jobs:
         run: dotnet test --no-build --verbosity normal --configuration Release --logger "trx;LogFileName=test-results.trx" --results-directory ./TestResults
 
       - name: Publish Test Results
+        # The docker-based action only runs on Linux; the Windows/macOS legs upload their .trx
+        # artifacts below and gate on the test step itself.
         uses: EnricoMi/publish-unit-test-result-action@v2
-        if: always()
+        if: always() && runner.os == 'Linux'
         with:
           files: 'TestResults/**/*.trx'
 
@@ -45,7 +54,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v7
         with:
-          name: test-results
+          name: test-results-${{ matrix.os }}
           path: TestResults/**/*.trx
 
   # mcp-authorize Phase-2 gate (design 04/07, task b8): the cross-language golden-vector parity
--- a/McpPlugin.Tests/AgentConfig/MachineCredentialStoreAtomicityTests.cs
+++ b/McpPlugin.Tests/AgentConfig/MachineCredentialStoreAtomicityTests.cs
@@ -1,0 +1,243 @@
+/*
+┌────────────────────────────────────────────────────────────────────────┐
+│  Author: Ivan Murzak (https://github.com/IvanMurzak)                   │
+│  Repository: GitHub (https://github.com/IvanMurzak/MCP-Plugin-dotnet)  │
+│  Copyright (c) 2025 Ivan Murzak                                        │
+│  Licensed under the Apache License, Version 2.0.                       │
+│  See the LICENSE file in the project root for more information.        │
+└────────────────────────────────────────────────────────────────────────┘
+*/
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using com.IvanMurzak.McpPlugin.AgentConfig;
+using Shouldly;
+using Xunit;
+
+namespace com.IvanMurzak.McpPlugin.AgentConfig.Tests
+{
+    /// <summary>
+    /// The atomic-write contract of the machine credential store (design 04 §1): temp sibling in
+    /// the same directory → fsync → rename, with the Windows rename retry policy (≥4 attempts,
+    /// ≥250 ms backoff on EBUSY/EPERM/EACCES-alike holders). A reader must never observe a
+    /// truncated document, and a writer killed before the rename must leave the destination intact.
+    /// </summary>
+    public sealed class MachineCredentialStoreAtomicityTests : IDisposable
+    {
+        private readonly string _baseDir;
+
+        public MachineCredentialStoreAtomicityTests()
+        {
+            _baseDir = Path.Combine(Path.GetTempPath(), "agd-store-atomic-" + Guid.NewGuid().ToString("N"), ".ai-game-dev");
+        }
+
+        public void Dispose()
+        {
+            var parent = Path.GetDirectoryName(_baseDir);
+            if (parent != null && Directory.Exists(parent))
+                Directory.Delete(parent, recursive: true);
+        }
+
+        private MachineCredentialStore NewStore() => new MachineCredentialStore(_baseDir);
+
+        private static MachineCredentials Credentials(string accessToken, string refreshToken) => new MachineCredentials
+        {
+            Families = new MachineCredentialFamilies
+            {
+                Plugin = new MachineCredentialFamily
+                {
+                    AccessToken = accessToken,
+                    RefreshToken = refreshToken,
+                    ClientId = "unity-mcp-plugin",
+                    Scope = "mcp:plugin",
+                },
+            },
+        };
+
+        [Fact]
+        public void RetryPolicy_MeetsTheStoreContract()
+        {
+            // Part of the v2 store contract (design 04 §1) — not a tunable.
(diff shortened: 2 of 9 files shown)


IvanMurzak:
## Taskflow review — depth **high** (blocking) — b1-cs-store-v2 @ b7005db

Reviewed against the immutable spec (design `04-credential-store-and-refresh.md` §1/§5, task `b1-cs-store-v2`), at depth high per the taskflow rubric (low DoD-conformance + medium reuse/test-meaningfulness/adjacent-call-sites + adversarial). **Result: no blocking findings. PASS.** Advisories below.

### DoD / spec conformance — verified

- **v2 model**: `families.{agent,plugin,legacy}` with per-family `clientId`+`scope`; `[JsonExtensionData]` at all three levels (top / families container / inside family) fixes the shipped drop-on-read. Serializer options are byte-identical to `main` (camelCase, case-insensitive, `WriteIndented`, null-omitting) — "unchanged from shipped" holds; tests compare parsed values, never bytes.
- **Atomic write**: temp sibling in the same directory (also the 0700 dir, which incidentally blocks temp-name symlink attacks), `Flush(flushToDisk:true)`, `File.Replace` (DACL-preserving) with `File.Move` create-fallback; the exists→Replace/Move race is closed by the retry loop re-evaluating (`FileNotFoundException` is an `IOException`, so both directions of the race retry). Retry = 5 attempts × 250 ms on `IOException`/`UnauthorizedAccessException` (EBUSY/EPERM/EACCES equivalents), pinned by a contract test (`≥4` / `≥250`). Advisory POSIX parent-dir fsync present, correctly fail-safe (all exceptions swallowed, incl. `DllNotFoundException`).
- **Mirror**: every write path funnels through `Write()` → `NormalizeForWrite()`; mirror re-stamped from `plugin ?? legacy`, **cleared** for agent-only documents (stale-mirror test plants a stale value and asserts absence). Old-v1-reader simulation asserts the raw top-level fields = the plugin family. No path writes a stale mirror (`WritePlaintextJson` bypasses, but it is `internal` and test-only).
- **v1 adoption / version**: `{version:1}` → `families.legacy` on read (version stays 1 in memory until first write — pinned); first write upgrades to v2+mirror preserving unknown fields; version > 2 passes through unmodified (pinned, incl. unknown-field survival); hybrid `{version:1, families}` is unrepresentable on write (`Version` is always lifted to ≥ 2).
- **Fail-closed**: `Rotate()` refuses (`InvalidOperationException`) on `Unreadable` and the file is byte-compared untouched; `Delete()` is reachable only from explicit sign-out (`SignOut()`); explicit-re-auth `Write()` replacing an unreadable store matches the design's stated boundary. `Exists` documented + tested as never a signed-in signal. The corrupted-fixture tests assert `Unreadable` **by enum value**, so they cannot pass via the `Missing` path.
- **Adjacent call sites** (every consumer of `Read()`/`Write()`/`Rotate()` in this repo): `PluginCredentialProvider` is the only one. Its `Rotate` call already catches and degrades to in-memory-only (no clobber); shipped `Read()` threw on unreadable and was caught → null, new `Read()` returns null directly → identical end state (`SignedOut` → recovery only via explicit sign-in → `Adopt`/`Write`, which is the permitted surface). `ServerTargetResolver` takes credentials by parameter only. No consumer sees changed semantics it doesn't handle.
- **Plants**: all four mandated plants exist and are structurally discriminating — (a) concurrent-reader hammer has a **real** anti-vacuity floor (`successfulReads ≥ 25` — if the window is never sampled the test fails rather than passing vacuously) and a generation-pairing assertion that catches torn documents, with 64 KB padding to widen the window; (b) the Windows handle-holder holds 600 ms against a 4×250 ms backoff budget (deterministically forces ≥ 2 backoffs, `elapsed ≥ 400 ms` floor proves the retry ran) — and windows-latest CI actually ran it; (c) corrupted store is a genuine non-DPAPI blob on Windows / non-JSON on POSIX, asserting `Unreadable`, never-delete byte-for-byte; (d) removing `[JsonExtensionData]` at any of the 3 levels reddens the round-trip test. PR body documents the RED mutation runs.
- **Scope**: no `ITokenRefresher` change, no lock protocol, `PluginCredentialProvider` untouched — no b2/b3 encroachment.
- **Workflow** (supply-chain check): no new third-party actions and no ref changes (`EnricoMi/publish-unit-test-result-action@v2` is pre-existing and now correctly Linux-gated — it is docker-based); per-OS artifact names de-collided; `fail-fast: false` correct for an OS-behavior matrix. Verified via the branch-protection API + ruleset 11607365 that **no named required status checks exist**, so the `test` → `test (<os>)` check-name change strands nothing. All 5 checks green on this head.

### ADVISORY (non-blocking)

1. **`Adopt()` silently depends on `Write()` mutating its argument.** `PluginCredentialProvider.Adopt` (`PluginCredentialProvider.cs:159-175`) decides `SignedIn` by reading `credentials.AccessToken` *after* `_store.Write(credentials)` — which only works for a families-only (no top-level tokens) credential because `NormalizeForWrite()` stamps the mirror **in place** on the caller's instance. Correct today, but the mutation contract isn't stated in `Write()`'s docs and no test pins Adopt-with-families-only → `SignedIn` (the only `Adopt` test uses top-level v1-style tokens). If `Write` ever normalizes a copy, families-only adoption becomes SignedOut-with-a-persisted-store. Suggest a one-line docblock note on `Write()` + a pinning test (natural home: b3 wave).
2. **POSIX temp-file permissions are ordering-guaranteed, not creation-guaranteed.** `WriteFileAtomic` creates the temp via `FileMode.CreateNew` (umask default, typically 0644) and chmods 0600 before any payload byte — correct, but the guarantee is a code-ordering invariant no test pins. On the net8/net9 TFMs, `FileStreamOptions.UnixCreateMode = 0600` would make it atomic at creation (keep the chmod path for netstandard2.1). Hardening only.
3. **`Rotate()` on a `Missing` store fabricates `families.legacy`.** Matches status-quo intent for old callers, but the b3 refresher must not reach family-aware writes through `Rotate()` (it can't address the agent family and invents `legacy` on an empty store). Boundary note for b3, consistent with the task split. Relatedly, `TryRead` classifies a persistent open failure (3×50 ms read-retry exhausted) as `Unreadable` — the safe direction (Rotate refuses) — but b3's refresher should treat that as "busy/retry", not "sign-out".
4. **Workflow least-privilege nit**: all three matrix legs inherit `checks: write` / `pull-requests: write`, though only the Linux leg's publish step needs them. Splitting the publish into its own job (or accepting the nit) is fine; `@v2` tag-pinning of the docker action is pre-existing and unchanged.
5. **OS-gated tests silently `return` (report "passed") off-OS** — e.g. the Windows-only handle-holder plant shows green on the POSIX legs. The matrix covers the real leg so coverage is intact, but a skip marker would make per-leg coverage legible instead of implying 3-OS coverage of a 1-OS test.

**Verdict: PASS** — high-depth review found no blocking issues; the five advisories above do not gate the merge.

PR #203, opened 2026-08-15
feat(auth): C# cross-process lock protocol for the machine credential store (b2-cs-lock)

IvanMurzak:
Implements taskflow `unified-machine-auth` task **b2-cs-lock** — design `04-credential-store-and-refresh.md` §2, byte-semantics twin of the upcoming TS lock (c2).

## What

`MachineCredentialLock` (`McpPlugin/src/AgentConfig/MachineCredentialLock.cs`):

- **Acquire**: `~/.ai-game-dev/credentials.lock`, `FileMode.CreateNew` **only** — no `FileShare.None`/flock reliance anywhere (O6). Writes `{pid, startedAt, hostId}`, closes the handle immediately so the mtime is peer-visible, jittered backoff on failure.
- **Constants** (shared cross-language contract, exported for x1 to assert against the TS values): `REFRESH_HTTP_TIMEOUT = 15_000` < `LOCK_STALE_MS = 60_000` < `ACQUIRE_BUDGET = 75_000` (ms; the ordering is the invariant), plus `FOREIGN_HOST_STALE_MS = 24 h`.
- **Stale takeover (compare-and-delete)**: last-**write**-time older than `LOCK_STALE_MS` AND local `hostId` → read → re-stat → delete only if byte-identical → race exclusive-create → **verify own content** before entering. Foreign or unparseable `hostId` → 24 h bar.
- **Busy**: budget exhaustion fails as "busy" (`TryAcquire()` → `null`) — never lock-free (D9 REVISED).
- **Guarded release**: never deletes a peer's lock (byte-compare before delete).
- **F6 logout delete path**: `TryDeleteStore(store)` = acquire → unlink store → release → unlink lock. `TryRunExclusive(Action)` is the seam b3 will wire `Rotate()` through (this PR does **not** touch `MachineCredentialStore` flows, per the task boundary).
- `hostId` = `Dns.GetHostName()` — same `gethostname` source as Node `os.hostname()`, so C# and TS processes on one machine see each other as local (c2 must mirror this).

## Deliberate decisions

- **`FileOptions.DeleteOnClose` (04 §2 "optional Windows hardening") omitted**: holding the handle conflicts with close-immediately mtime visibility and diverges observable semantics from the TS twin; crash release is provided by the stale-takeover path (proven by plant 1). Documented in the class docs.
- **Unparseable lock document** (e.g. empty file from a writer killed between create and write) has no provable local owner → conservative 24 h foreign bar, never the 60 s local bar. c2 should mirror.

## Tests (real subprocesses for the takeover plants — 04 §5)

New `McpPlugin.Tests.LockHarness` console project (test-only, `IsPackable=false`), spawned as real OS processes with the **real contract constants**:

1. **kill -9 recovery**: holder killed mid-hold → survivor recovers after `LOCK_STALE_MS` (lower time bound asserted against the dead lock mtime) within `ACQUIRE_BUDGET`.
2. **live holder**: holder sleeping `REFRESH_HTTP_TIMEOUT` inside the lock is never taken over (holder itself re-verifies byte-identity → `OWNED=true`); the concurrent waiter enters only after the release.
3. **two waiters, one stale lock**: barrier-synchronized start; exclusive critical-section entry proven by an exclusive-create sentinel + sequential-timing assertion; 3 rounds.

Plus 16 in-process tests: constant values + ordering invariant, document semantics, busy-never-lock-free, foreign/unparseable 24 h bar, guarded release, idempotent release, `TryRunExclusive`, logout delete path (incl. busy → store intact).

## RED verification (G-SEC-1) — each plant fails when its control is removed

| Mutation | Plant | Observed RED |
|---|---|---|
| M1: stale takeover call removed | 1 | survivor `BUSY elapsed=75006`, exit 2 |
| M2: local stale bar → 5 s (breaks `REFRESH_HTTP_TIMEOUT < LOCK_STALE_MS`) | 2 | holder `OWNED=false`, exit 3 |
| M3: compare-and-delete + verify-own removed | 3 | both waiters ACQUIRED 4 ms apart, `OVERLAP` on the sentinel |

All mutations reverted; full solution suite green locally on net8.0 + net9.0 (851/851 McpPlugin.Tests, 684/684 McpPlugin.Server.Tests).

### DIFF
--- a/McpPlugin.Tests.LockHarness/McpPlugin.Tests.LockHarness.csproj
+++ b/McpPlugin.Tests.LockHarness/McpPlugin.Tests.LockHarness.csproj
@@ -1,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!-- Real-subprocess harness for the machine credential lock takeover plants (design 04 §5:
+       "Concurrency - real processes, not threads"). Referenced by McpPlugin.Tests, which spawns
+       it as an actual OS process to exercise kill -9 recovery, holder liveness, and the
+       two-waiter compare-and-delete race. Test infrastructure only - never packed. -->
+
+  <PropertyGroup>
+    <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <LangVersion>11.0</LangVersion>
+    <IsPackable>false</IsPackable>
+    <RootNamespace>com.IvanMurzak.McpPlugin.Tests.LockHarness</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../McpPlugin/McpPlugin.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/McpPlugin.Tests.LockHarness/Program.cs
+++ b/McpPlugin.Tests.LockHarness/Program.cs
@@ -1,0 +1,169 @@
+/*
+┌────────────────────────────────────────────────────────────────────────┐
+│  Author: Ivan Murzak (https://github.com/IvanMurzak)                   │
+│  Repository: GitHub (https://github.com/IvanMurzak/MCP-Plugin-dotnet)  │
+│  Copyright (c) 2025 Ivan Murzak                                        │
+│  Licensed under the Apache License, Version 2.0.                       │
+│  See the LICENSE file in the project root for more information.        │
+└────────────────────────────────────────────────────────────────────────┘
+*/
+
+#nullable enable
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using com.IvanMurzak.McpPlugin.AgentConfig;
+
+namespace com.IvanMurzak.McpPlugin.Tests.LockHarness
+{
+    /// <summary>
+    /// Subprocess harness for the machine-credential-lock takeover plants (design 04 §5 — real OS
+    /// processes, not threads). Spawned by <c>MachineCredentialLockTakeoverPlantTests</c>. Every
+    /// protocol-relevant event is reported as a machine-parseable stdout line stamped with a
+    /// wall-clock unix-ms timestamp, flushed immediately so the parent can react (e.g. kill -9 a
+    /// holder the moment it reports <c>ACQUIRED</c>).
+    ///
+    /// <para>Commands (all use the PUBLIC lock constructor — real contract constants):</para>
+    /// <list type="bullet">
+    ///   <item><c>hold &lt;baseDir&gt; &lt;holdMs&gt;</c> — acquire, report, sleep
+    ///         <c>holdMs</c> inside the critical section, verify the lock is still our own
+    ///         (<c>OWNED=…</c>), release. Exit 0 = held throughout; 2 = busy; 3 = stolen.</item>
+    ///   <item><c>acquire &lt;baseDir&gt; &lt;csHoldMs&gt; [--wait-go]</c> — acquire within the
+    ///         budget, prove exclusive entry via an exclusive-create sentinel
+    ///         (<c>cs.sentinel</c>), hold the critical section <c>csHoldMs</c>, release. Exit
+    ///         0 = clean exclusive run; 2 = busy; 4 = OVERLAP (a second holder was inside the
+    ///         critical section — mutual-exclusion violation). With <c>--wait-go</c> the process
+    ///         reports <c>READY</c> and spins until <c>go</c> appears in <c>baseDir</c> before
+    ///         acquiring — a start barrier so two waiters observe the SAME stale lock instead of
+    ///         being serialised by process-spawn jitter.</item>
+    /// </list>
+    /// </summary>
+    internal static class Program
+    {
+        private const int ExitOk = 0;
+        private const int ExitUsage = 1;
+        private const int ExitBusy = 2;
+        private const int ExitStolen = 3;
+        private const int ExitOverlap = 4;
+
+        /// <summary>Sentinel proving exclusive critical-section entry in the two-waiter plant.</summary>
+        private const string SentinelFileName = "cs.sentinel";
+
+        private static long NowUnixMs => DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+
+        private static void Report(string line)
+        {
+            Console.Out.WriteLine(line);
+            Console.Out.Flush(); // the parent test reacts to lines in real time — never buffer
+        }
+
+        private static int Main(string[] args)
+        {
+            if (args.Length < 3 || args.Length > 4)
+            {
+                Console.Error.WriteLine("usage: (hold|acquire) <baseDir> <milliseconds> [--wait-go]");
+                return ExitUsage;
+            }
+
+            var command = args[0];
+            var baseDir = args[1];
+            var milliseconds = int.Parse(args[2]);
+            var waitGo = args.Length == 4 && args[3] == "--wait-go";
+
+            switch (command)
+            {
+                case "hold":
+                    return Hold(baseDir, milliseconds);
+                case "acquire":
+                    return Acquire(baseDir, milliseconds, waitGo);
+                default:
+                    Console.Error.WriteLine("unknown command: " + command);
+                    return ExitUsage;
+            }
+        }
+
+        /// <summary>
+        /// Start barrier: report <c>READY</c>, then spin until the parent creates
+        /// <c>&lt;baseDir&gt;/go</c>. Lets the parent line up N waiters and release them within a
+        /// few milliseconds of each other.
+        /// </summary>
+        private static void AwaitGoSignal(string baseDir)
+        {
+            Report("READY t=" + NowUnixMs);
+            var goPath = Path.Combine(baseDir, "go");
+            var deadline = DateTime.UtcNow + TimeSpan.FromSeconds(30);
+            while (!File.Exists(goPath))
+            {
+                if (DateTime.UtcNow > deadline)
+                    throw new TimeoutException("go signal never arrived at " + goPath);
+                Thread.Sleep(5);
+            }
+        }
+
+        private static int Hold(string baseDir, int holdMs)
+        {
+            var mutex = new MachineCredentialLock(baseDir);
+            var stopwatch = Stopwatch.StartNew();
+            var handle = mutex.TryAcquire();
+            if (handle == null)
+            {
+                Report("BUSY elapsed=" + stopwatch.ElapsedMilliseconds);
+                return ExitBusy;
+            }
+
+            Report("ACQUIRED t=" + NowUnixMs + " elapsed=" + stopwatch.ElapsedMilliseconds + " pid=" + Environment.ProcessId);
+
+            Thread.Sleep(holdMs); // the "refresh in flight" window — a kill -9 lands here
+
+            var owned = handle.IsStillOwned();
+            Report("OWNED=" + (owned ? "true" : "false") + " t=" + NowUnixMs);
+
+            handle.Dispose();
+            Report("RELEASED t=" + NowUnixMs);
+            return owned ? ExitOk : ExitStolen;
+        }
+
+        private static int Acquire(string baseDir, int csHoldMs, bool waitGo)
+        {
+            if (waitGo)
+                AwaitGoSignal(baseDir);
+
+            var mutex = new MachineCredentialLock(baseDir);
+            var stopwatch = Stopwatch.StartNew();
+            var handle = mutex.TryAcquire();
+            if (handle == null)
+            {
+                Report("BUSY elapsed=" + stopwatch.ElapsedMilliseconds);
+                return ExitBusy;
+            }
+
+            var acquiredAt = NowUnixMs;
+            Report("ACQUIRED t=" + acquiredAt + " elapsed=" + stopwatch.ElapsedMilliseconds + " pid=" + Environment.ProcessId);
+
+            // Exclusive-create sentinel: if a second process is inside the critical section at the
+            // same time, its CreateNew fails and it reports OVERLAP — the direct, cross-process
+            // proof of the mutual-exclusion property the two-waiter plant asserts.
+            var sentinelPath = Path.Combine(baseDir, SentinelFileName);
+            try
+            {
+                using (new FileStream(sentinelPath, FileMode.CreateNew, FileAccess.Write, FileShare.Read))
+                {
+                }
+            }
+            catch (IOException)
+            {
+                Report("OVERLAP t=" + NowUnixMs);
+                handle.Dispose();
+                return ExitOverlap;
+            }
+
+            Thread.Sleep(csHoldMs); // widen the critical-section window so an overlap is observable
+
+            File.Delete(sentinelPath);
+            handle.Dispose();
+            Report("DONE t=" + NowUnixMs);
+            return ExitOk;
+        }
+    }
+}
--- a/McpPlugin.Tests/AgentConfig/MachineCredentialLockTakeoverPlantTests.cs
+++ b/McpPlugin.Tests/AgentConfig/MachineCredentialLockTakeoverPlantTests.cs
@@ -1,0 +1,351 @@
+/*
+┌────────────────────────────────────────────────────────────────────────┐
+│  Author: Ivan Murzak (https://github.com/IvanMurzak)                   │
+│  Repository: GitHub (https://github.com/IvanMurzak/MCP-Plugin-dotnet)  │
+│  Copyright (c) 2025 Ivan Murzak                                        │
+│  Licensed under the Apache License, Version 2.0.                       │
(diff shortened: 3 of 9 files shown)


IvanMurzak:
## Taskflow review — depth **high** (blocking) — b2-cs-lock @ `2ba0261`

Reviewed against 04 §2/§5 (immutable spec) per the taskflow rubric (low+medium+high, strictly additive). Conformance verified first: exclusive-create only; no `FileShare.None`/flock reliance for mutual exclusion (sharing flags are used solely to avoid blocking peer reads/deletes — in-spec); `{pid, startedAt, hostId}` content; close-immediately (tested); jittered backoff; constants exported + ordering pinned; compare-and-delete takeover with verify-own; foreign 24 h bar; busy-never-lock-free with **no kill-switch reintroduction anywhere** (grepped); F6 logout order correct; `MachineCredentialStore` untouched (b3 boundary OK); harness `IsPackable=false` and referenced only by the non-packable test project (no NuGet contamination); netstandard2.1 leg clean (`Process.GetCurrentProcess().Id` not `Environment.ProcessId`; System.Text.Json + chmod patterns identical to the store's existing use in the same assembly/TFMs). All three plants use the real contract constants via the public ctor, and each RED mutation discriminates the property it claims.

Two findings are **BLOCKING**; the rest are advisory.

---

### BLOCKING

**B1 — `TryCreateOwnLock` leaks a poisoned empty lock file when the write fails after a successful `CreateNew`** (`MachineCredentialLock.cs:249-267`).

If the `FileStream` ctor succeeds but `stream.Write` (or the flush in the `using` dispose) throws — disk full, quota, I/O error, AV `UnauthorizedAccessException` — the `catch` returns `null` and **the empty file this process just created is left behind**. Every peer, *and this process's own next retry iteration*, then classifies it unparseable → foreign → **24 h bar** (`:296-299`, `:384-388`). Because D9 forbids proceeding lock-free and login/logout are also lock-gated (`TryDeleteStore`, the two-hold login path), one transient disk-full event freezes **all** credential operations on the machine for 24 hours — persisting after space is freed, with no product surface that deletes the artifact.

Fix: the ctor succeeding proves this process owns the file, so distinguish create-failure from write-failure and best-effort `File.Delete(_lockPath)` before returning `null` on the post-create path. This is race-safe: a milliseconds-old lock can never be taken over, and exclusive-create excludes peers, so the path can only hold our own artifact at cleanup time.

**B2 — the derived decision "unparseable/empty lock doc ⇒ 24 h foreign bar" is unsound in the availability direction** (`:296-299`, `ParseHostId` `:372-390`; pinned by `UnparseableLock_IsTreatedAsForeign_NotTakenOverAtTheLocalBar`).

An unparseable document's writer **provably never entered the critical section**: the full-document write precedes handle-return, which precedes entry. So taking over an unparseable lock can never collide with an in-flight refresh — the cross-host clock-skew rationale for the 24 h bar protects nothing for this class. The cost side: any empty/corrupt artifact (kill/power-loss in the create→write window; B1's disk-full leak) produces a 24 h machine-wide auth freeze with no self-heal — defeating the stated purpose of `LOCK_STALE_MS < ACQUIRE_BUDGET` ("takeover is reachable"), which fails outright for this reachable crash class. The frozen-writer edge (a writer suspended between create and write, waking after takeover to write into an unlinked inode on POSIX and entering — the normal create path has no verify-own) exists at *any* threshold including 24 h, so the long bar does not close it either; it only relocates the probability.

Recommend: unparseable ⇒ **local 60 s bar** (the writer-never-entered argument), updating the pinning test; and either way, record the resolved decision as an explicit **c2-mirror + x1 contract item** — the two implementations diverging here would make one side's crash artifact block the other side for a day. If 24 h is deliberately kept, B1 must still be fixed and the residual (kill-in-window ⇒ 24 h outage) explicitly accepted on this PR.

---

### ADVISORY

**A1 — residual takeover TOCTOU: judged to match the spec-as-written, not exceed it.** `File.Delete` at `:319` is delete-by-path; between the re-read (`:313`) and the delete, preemption/laptop-sleep makes the window unbounded, so a waiter can delete an arbitrary *later* generation (including a live new holder's fresh lock), and a both-enter interleaving exists when the victim's verify-own completes before the thief's delete. POSIX/NTFS have no compare-and-delete primitive and 04 §2 mandates exactly this sequence with verify-own as the sole narrowing — the residual is the spec's own, and "absorbed by the grace window" (D10, 30 s idempotent replay) is the design's stated mitigation. Actionable hardening for **b3**: call `handle.IsStillOwned()` immediately before the store write inside the critical section (the handle already exposes it; plant 2's holder demonstrates the pattern) so a rare both-enter degrades to a detected abort instead of a lost update. Re-stat freshness confirmed sound: static `File.GetLastWriteTimeUtc` stats on every call (no `FileInfo` caching).

**A2 — verify-own is blind to byte-identical documents from the same process; material for c2/x1.** Two concurrent takeover flows in one process building `{pid, startedAt, hostId}` in the same clock tick produce identical bytes: flow B can delete flow A's fresh lock, recreate an identical document, and both verify "own". In C# `startedAt` is 100 ns-granular (precise system time), so odds are negligible — advisory here. But Node's `toISOString()` is **millisecond**-granular and two async acquire flows in one event loop can plausibly tie: **c2 should add a per-attempt nonce/high-resolution component or intra-process serialization, and x1 should pin the chosen disambiguator.** A static per-path intra-process gate on the C# side would close it for free.

**A3 — clock/timebase, all spec-inherent, no change requested.** Backward clock jump ⇒ negative age ⇒ takeover merely delayed (safe direction). Forward step > ~45 s (NTP step, resume-from-sleep RTC correction) ⇒ a live holder can be declared stale prematurely ⇒ double refresh ⇒ D10 absorbs. NFS-hosted home: mtime is server-clock vs local `UtcNow` — a >60 s skew makes a fresh local lock instantly stale. Worth one doc line; c2 will be identical. Also: `GetLastWriteTimeUtc` on a vanished file returns the 1601 FILETIME-zero sentinel rather than throwing — the path is safe (the re-read null-check plus `startedAt` uniqueness close every interleaving I could construct), but a comment would spare the next reader the same derivation.

**A4 — hostId contract items to pin in x1 / mirror in c2**: (a) source `Dns.GetHostName()` = Node `os.hostname()`; (b) the failure fallback `Environment.MachineName` (NetBIOS: upper-cased, 15-char truncated) can diverge from a TS peer's hostname → mismatch → foreign bar — over-conservative direction only, never wrongful takeover; (c) comparison is `OrdinalIgnoreCase` — **the TS twin must also compare case-insensitively**; (d) empty-string hostname is self-consistent.

**A5 — plant quality: all three non-vacuous.** The OVERLAP sentinel is a genuine cross-process exclusive-create witness; its one blind sliver (between sentinel-delete and lock-release, microseconds) does not cover the targeted interleavings, which land both entries within ms (observed RED at 4 ms under M3). `DisableParallelization` is a documented test-infra measure (CPU starvation of an unrelated timing test) — the plants' own concurrency is cross-subprocess and unaffected, so no product race is masked; the timing-override ctor overrides only the budget in-process while staleness constants stay real, and the real-constants public ctor is exercised by all three plants. One flake risk: plant 2's anti-vacuity bound `elapsed >= 8_000` (`MachineCredentialLockTakeoverPlantTests.cs:160`) assumes waiter spawn+JIT stays under ~7 s; a cold 2-core runner could false-RED — consider deriving the bound from the holder's ACQUIRED/RELEASED timestamps instead.

**A6 — minor reuse note**: `SetPosixDirectoryPermissions` duplicates the store's private helper — justified by the b2 boundary and documented; consolidate when b3 opens that file.

---

**Verdict: BLOCK** — B1 and B2 go back to the implementer under the fix loop (K = 2); everything else is advisory and does not gate.

IvanMurzak:
## Taskflow re-review — fix round 1 (depth **high**) — b2-cs-lock @ `c2388a1`

Re-reviewed the actual delta `2ba0261..c2388a1` (3 commits) against the owner-adopted cross-language amendment, with the merged TS twin (cli-core `src/credential-lock.ts` + `test/golden-vectors/LockProtocol.GoldenVectors.json` at `c1d2cc0` on main) as the mirror reference. CI 5/5 green at this head (run 31874186118, incl. the `golden-vector-parity` job).

### Blocking findings from round 0 — both CLOSED

**B1 (poisoned-artifact leak) — closed.** Both create sites now go through one `TryExclusiveCreateWithContent` (`MachineCredentialLock.cs:369-391`) that tracks `created` and best-effort-deletes the file this process provably owns when the write fails after a successful `CreateNew` — for the lock AND the intent file, mirroring the TS twin's cleanup blocks. The `WriteFaultInjection` seam is verified test-only: an `internal` field, `null` by default, invoked via `?.Invoke()` inside the create window, `InternalsVisibleTo` limited to `McpPlugin.Tests` (the LockHarness sees public API only) — zero production behavior change when unset. Pinned deterministically by three tests, each judged non-vacuous: persistent fault ⇒ busy AND no artifact survives (`CreateWriteFailure_NeverLeavesAPoisonedLockArtifact` — RED without the cleanup, the empty file would exist); one-shot fault ⇒ next attempt acquires with a complete document (`faults >= 2` anti-vacuity); same control at the intent site, which additionally proves the stale lock is never deleted without an owned intent.

**B2 (24 h bar for unparseable docs) — closed per the amendment.** `ClassifyLockDocument` (`:657-686`) mirrors TS `classifyLockDocument` branch-for-branch — I compared them case-by-case: zero-byte / invalid JSON / valid-but-non-object (number, array, null, string) ⇒ `Unparseable` ⇒ 60 s; JSON object with missing/empty/non-string/different `hostId` ⇒ `Foreign` ⇒ 24 h; `OrdinalIgnoreCase` local match (TS `toLowerCase()` — divergent only on non-ASCII hostnames, in the conservative direction, documented). Applied to the lock AND the intent. The old 24 h pin test is flipped (`UnparseableLock_IsStaleEligibleAtTheLocalBar_AndEmitsOneDiagnostic`), the not-before-60 s half is separately pinned (`FreshUnparseableLock_IsNotTakenOverBeforeTheLocalBar`), the intent-side half is pinned (`ZeroByteIntent_DoesNotWedgeTakeoverFor24Hours`), and the diagnostic asymmetry matches the TS twin exactly: one warning after an unparseable LOCK removal, intent recovery silent (asserted).

### Arbiter mirror fidelity — verified step-for-step against the merged TS

I walked `TryRemoveStaleLock`/`TryAcquireTakeoverIntent` against TS `tryStaleTakeover`/`tryAcquireTakeoverIntent`: stat1(mtime+size)+read1 → classify → threshold → intent exclusive-create → intent read-back verify → candidate re-validate (mtime + size + byte-identical) → pre-removal intent ownership re-verify → delete → unparseable diagnostic after removal → intent release in `finally`, guarded by content compare; crashed-intent recovery via the same classifier + compare-and-delete, deliberately unserialized in both languages with the same documented residual (crashed-intent precondition × tight window); live intent ⇒ back off. Windows delete-pending treated as contention at both create sites (C# necessarily folds it into the IOException/UAE catch; TS enumerates EPERM/EACCES/EBUSY — same observable outcome). Normal-path acquire now also read-back-verifies its own content, closing round 0's frozen-writer note. Protocol-level semantics: **equal**.

The three deterministic pins discriminate the right controls — I verified each mutation flips its named test *by construction*, not just by report:
- **M-ARB** (`StaleTakeover_BacksOff_WhileALiveTakeoverIntentExists`): bypassing the arbiter takes over the stale lock despite a live claimant's intent ⇒ acquire succeeds ⇒ `ShouldBeNull()` RED; with the control, busy with both artifacts byte-untouched.
- **M-RB** (`IntentReadBack_DetectsADisplacedIntent_BeforeRelyingOnIt`): the displace-then-restore choreography is the right one — restoring our bytes at the later seam makes the pre-removal re-verify pass, so ONLY the read-back can catch the displacement; without it the takeover completes ⇒ RED. Anti-vacuity (`ourIntentBytes` captured) holds.
- **M-RV** (`PreRemovalIntentReVerify_AbortsWhenTheArbiterIsDisplaced_AfterAcquisition`): displacement after acquisition is invisible to the read-back; without the re-verify the claimant deletes the lock under a stolen arbiter and acquires ⇒ RED. With it: busy, lock untouched, displaced intent respected.

### Nonce / hostId / vector parity

`{pid, startedAt, hostId, nonce}` with a fresh 128-bit lowercase-hex nonce per acquisition (`RandomNumberGenerator`, netstandard2.1-safe), serialization order = declaration order = the vector's `lockDocumentFields` — asserted against the ACTUAL written document (`Constants_MatchTheCommittedGoldenVector`). I `cmp`'d the committed C# vector against cli-core main's: **byte-identical**. `hostIdCompare` pinned; `OrdinalIgnoreCase` in code + case-insensitivity test; empty/missing/non-string `hostId` in a well-formed doc ⇒ Foreign (tested); nonce-freshness across same-instance acquisitions tested (round-0 A2 resolved in both languages — TS carries the same field, optional on read for pre-nonce writers).

### The out-of-diff commit `c2388a1` — accepted

`Write_RetriesUntilDestinationHandleReleased_Windows`: `Task.Run` → dedicated `Thread` (+ `async Task` → `void`/`Join`). All assertions unchanged (400 ms elapsed floor, content, teardown). The rationale — thread-pool delay under the plant collection's subprocess CPU bursts postponing the 600 ms release past the retry window — is consistent with the starvation interaction the plant collection's own doc named in round 0, with this exact test as the victim. Justified test-only stabilization; no revert demanded.

### Plants, boundary, packaging

The three mandated subprocess plants and the harness are untouched by the delta and drive the real public-ctor code, so they now exercise the intent-serialized mechanism end-to-end with the real constants — I re-walked all three against the new mechanism (plant 3's two waiters now serialize through the intent; OVERLAP and the sequential-timing assertion remain valid witnesses). `MachineCredentialStore` remains untouched (delta = 5 files, none of them the store). `LockProtocol.GoldenVectors.json` sits under `McpPlugin/src/` as a default `None` item — NOT packed (only README/LICENSE carry `Pack="true"`), following the existing ProjectIdentity-vector precedent, linked into the test output the same way.

### Round-0 advisories — status

- **A1 (takeover TOCTOU)**: hardened beyond round 0 — the intent arbiter deterministically closes the two-waiter delete-wrong-generation interleaving; residual = crashed-intent recovery race, accepted by design in both languages. The `IsStillOwned()`-before-store-write guidance for b3 is now in the handle's doc. Resolved.
- **A2 (byte-identical documents)**: resolved via the nonce, both languages.
- **A3 (clock/timebase + 1601 sentinel)**: resolved by documentation in the class doc.
- **A4 (hostId contract)**: resolved — pinned in the vector + tests; `MachineName` fallback divergence documented as conservative-direction-only.
- **A5 (plant 2's `elapsed >= 8_000` cold-runner flake risk)**: unchanged — stands as an advisory.
- **A6 (chmod helper duplication)**: unchanged — consolidate when b3 opens the store file.

### New ADVISORY findings (round 1) — none block

- **R1-A1 — the PR body is stale and now contradicts the shipped contract.** It still documents the round-0 protocol: "Foreign or unparseable `hostId` → 24 h bar", a `{pid, startedAt, hostId}` document without `nonce`, no intent file, and only the M1-M3 plant mutations (M-ARB/M-RB/M-RV are not on the PR record at all). A merged PR whose description asserts the opposite of the adopted amendment on exactly the amended points is a citation hazard for c2/x1 readers. Please update the body before merge.
- **R1-A2 — error-surfacing divergence (deliberate-looking, worth a doc line for b3/c3)**: TS `acquire()` throws the environmental error (e.g. ENOSPC) out of a failed write immediately; C# folds every create/write `IOException`/`UnauthorizedAccessException` into contention and reports "busy" after the 75 s budget (it also cannot distinguish disk-full from Windows delete-pending, unlike Node's errno). On-disk protocol identical; per-language caller experience differs.
- **R1-A3 — no intra-process reentrancy guard in C#**: TS throws on `acquire()` of a held instance; a C# caller that re-enters `TryAcquire` while holding burns the budget and gets "busy". The handle-based API makes this less meaningful and b3's per-process single-flight (§3.6) mitigates; noting for the b3 wiring.
- **R1-A4 — TS-side doc rot (cli-core, out of this repo's scope)**: `tryStaleTakeover`'s method doc (credential-lock.ts ~:422-425) still says "Unparseable content is treated as FOREIGN on purpose", contradicting its own classifier and the adopted amendment. Flagging so x1/c2 don't cite the stale sentence as the contract.

---

**Verdict: PASS** — both round-0 blocking findings are resolved and verified against the actual diff; the amendment is mirrored faithfully; no new blocking findings. Advisories above do not gate.

IvanMurzak:
Record note (orchestrator): the PR body describes the round-0 protocol. The FINAL protocol at head c2388a1 is the owner-adopted cross-language amendment — intent-file takeover serialization (credentials.lock.takeover), unparseable/zero-byte ⇒ 60s stale-eligibility (lock+intent, one post-removal diagnostic on lock side only), delete-pending=contention, {pid,startedAt,hostId,nonce} doc, OrdinalIgnoreCase hostId. Authoritative G-SEC-1 RED/GREEN mutation record for the final controls (M-B1/M-B2/M-B2i/M-ARB/M-RB/M-RV): see the fix-round report and re-review at issuecomment-5301417427. TS twin: AI-Game-Dev-CLI-Core#8 (merged c1d2cc0).

PR #204, opened 2026-08-15
test(auth): x1 cross-language golden-vector parity — store v2, cohort, DPAPI cross-impl

IvanMurzak:
## Summary

C# twin of the cli-core PR (unified-machine-auth 04 §1, task **x1-golden-parity**). Companion PR:
IvanMurzak/AI-Game-Dev-CLI-Core#9. Adds the store v2 cross-language golden-vector consumption this
repo was missing (only `LockProtocol`/`ProjectIdentity` vectors existed here before).

- `McpPlugin/src/AgentConfig/MachineCredentials.GoldenVectors.json`: byte-identical copy of
  `cli-core/test/golden-vectors/MachineCredentials.GoldenVectors.json` (verified with `diff` +
  SHA256), wired into `McpPlugin.Tests.csproj` the same way `LockProtocol`/`ProjectIdentity`
  vectors already are.
- `MachineCredentialsGoldenVectorTests.cs`: v2 round-trip, v2→v1-mirror old-reader simulation
  (non-vacuous — strips the vector's pre-baked top-level triple before `Write()` so the mirror step
  is the only source of truth; verified RED locally against a mandated mirror-drop plant in
  `MachineCredentials.NormalizeForWrite`, then reverted), v1→v2 adoption matching the golden
  adoption result exactly.
- `MachineCredentialsCohortTests.cs`: the two 06-migration-rollout cohort rows with a store
  file-shape — v1-only reader/legacy family (row 1) and mixed writer/old reader (row 4, using the
  new `cohortMixedVersionsOldWriterCollapse` vector) — "no cohort regresses below status quo" (O4).
- `MachineCredentialDpapiCrossImplementationTests.cs` (Windows CI leg): round-trips a blob
  encrypted via the store's real `CryptProtectData` P/Invoke (new `ProtectBytes` test seam,
  mirroring the existing `UnprotectBytes`) through a literal transcription of cli-core's
  `dpapiTransform` PowerShell script, and vice versa. Includes the mandated entropy plant.
- `MachineCredentialStore.cs`: adds the `ProtectBytes` internal test seam (`InternalsVisibleTo:
  McpPlugin.Tests`) used by the DPAPI cross-implementation tests above.

## Test plan

- [x] `dotnet build McpPlugin.Tests/McpPlugin.Tests.csproj` (net8.0 + net9.0) — clean
- [x] `dotnet test` (net9.0) — 877 passed, 0 failed
- [x] `dotnet test` (net8.0, new test files) — 9 passed, 0 failed
[...]

PR #205, opened 2026-08-15
fix(network): close ConnectionManager single-flight race + de-flake rejection tests

IvanMurzak:
## Root cause of the ConnectionManager flaky-test family

~9 documented red CI events over one day across ubuntu/macOS/windows, on PRs whose diffs never touch Network code (PR #203 runs + 4 consecutive rerun attempts of PR #204's run 31876647434). Two distinct mechanisms, both diagnosed by **forcing them deterministically**:

### 1. Product race — `Connect_ThreadSafety_NoRaceConditionsUnderLoad` (`callCount` 1 vs 2)

`ConnectionManager.Connect`'s single-flight was a **non-atomic check-then-act**: the `_ongoingConnectionTask` null-check and the publication of the attempt task were separated by the `_gate` acquisition. A caller that read `null` in that window queued on `_gate` — which the leader holds for the **whole** attempt — woke only after the attempt completed and the field was cleared, and then started a second, duplicate connection attempt (a second `CreateConnectionAsync` call). The window is microseconds wide, which is why it only fired on loaded CI runners.

**Deterministic proof:** the new pin test `Connect_CallerInPrePublicationWindow_JoinsInFlightAttempt` issues a second `Connect` synchronously from a `KeepConnected` subscription — R3 fires it on the leader's thread exactly inside the race window (Connect enables reconnection before publishing). On the old code it fails with the exact CI shape, no sleeps involved:

```
Shouldly.ShouldAssertException : providerCallCount  should be 1  but was 2
```

**Fix:** `Connect` now atomically either JOINS the in-flight attempt or INSTALLS itself as leader under one `_ongoingConnectionGate` hold, publishing a `TaskCompletionSource` proxy (the real task can only be created under `_gate`, and `_gate` must not be taken while holding `_ongoingConnectionGate` — `Disconnect` acquires them in the opposite order, so doing that would deadlock). The leader's cleanup clears the field only if it still owns it (`DisconnectImmediate` may have detached the attempt and a successor may have installed itself) and always completes the proxy so joined callers are never orphaned. The now-dead in-gate double-check is removed.

This also hardens the sibling assertions (`Connect_WhenMultipleThreadsCallSimultaneously_ProviderCalledOnlyOnce`, `Connect_WhenCalledConcurrently_ReusesFirstConnectionAttempt`, `Disconnect_WhenCalledAfterConcurrentConnects_PreventsQueuedConnectFromProceeding`) which assert the same guaranteed-coalescing contract and carried the same latent race.

### 2. Test time-budget — `Connect_FailedAttemptsResetRejectionCounter` (windows, preceded by parallel-fixture log spam)

The loop self-terminates via sequence exhaustion; the 5s CTS was meant as a hang guard but silently doubled as a **timing assumption**: the happy path already spends ~1.2s of real time (the product's hard 1s post-create stabilization delay + 3×50ms rejection windows), so runner load stretching timer resumptions past the remaining budget truncated the loop mid-sequence.

**Deterministic proof:** shrinking the budget to 1.1s reproduces the exact CI shape:

```
Shouldly.ShouldAssertException : cm.AttemptCount  should be greater than 3  but was 3
```

**Fix:** in both self-terminating rejection tests the CTS is now a pure 30s hang guard with an explicit `cts.Token.IsCancellationRequested.ShouldBeFalse(...)` assertion that it never fires — the pattern this file already uses in `Connect_StopsAfterConsecutiveConnectionFailures_WhenOptedIn`. No assertion was weakened; each test gained one.

## Evidence

| Check | Result |
|---|---|
| Pin test on old code | RED — `providerCallCount should be 1 but was 2` (CI shape) |
| Rejection test with squeezed budget on old code | RED — `AttemptCount should be greater than 3 but was 3` (CI shape) |
| Pin test on fixed code | GREEN, 25/25 repeat runs |
| 100-iteration repeat runs (2 flaky tests + pin test), net8.0 + net9.0, under concurrent full-suite load | **0 failures** on both TFMs |
| Baseline 100-iteration runs on unfixed code (idle dev box) | 0 failures — consistent with the race being load-dependent, hence the deterministic forcing above |
| Full `McpPlugin.Tests` suite on the merged tree | **880/880 green on net8.0 and net9.0** |

### DIFF
--- a/McpPlugin.Tests/Network/Connection/ConnectionManagerRejectionTests.cs
+++ b/McpPlugin.Tests/Network/Connection/ConnectionManagerRejectionTests.cs
@@ -55,11 +55,18 @@ namespace com.IvanMurzak.McpPlugin.Tests.Network.Connection
                 _logger, _testVersion, _testEndpoint, _mockProvider.Object
             );
 
-            // Act
-            using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
+            // Act: the loop terminates ON ITS OWN via the rejection cap; the CTS is a pure hang
+            // guard that must NEVER fire. It needs generous headroom: the happy path already
+            // spends real time (ConnectionManager's 1s post-create stabilization delay plus one
+            // RejectionThreshold window per rejection), and on a loaded CI runner every timer
+            // resumption can stretch by seconds — a tight budget turns runner load into a
+            // truncated loop and a bogus AttemptCount failure (the documented CI flake).
+            using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
             var result = await cm.Connect(cts.Token);
 
             // Assert
+            cts.Token.IsCancellationRequested.ShouldBeFalse(
+                "The loop must stop via the rejection cap, not the hang-guard CTS");
             result.ShouldBeFalse("Connection should fail after repeated rejections");
             cm.KeepConnected.CurrentValue.ShouldBeFalse("KeepConnected should be disabled after rejection threshold");
             cm.AttemptCount.ShouldBe(3, "Should have attempted exactly MaxConsecutiveRejections times");
@@ -80,12 +87,20 @@ namespace com.IvanMurzak.McpPlugin.Tests.Network.Connection
                 attemptResults: sequence
             );
 
-            // Act
-            using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
+            // Act: the loop terminates ON ITS OWN when the sequence is exhausted (the manager
+            // flips _continueToReconnect off); the CTS is a pure hang guard that must NEVER
+            // fire. It needs generous headroom: the happy path already spends real time
+            // (ConnectionManager's 1s post-create stabilization delay plus one
+            // RejectionThreshold window per rejection), and on a loaded CI runner every timer
+            // resumption can stretch by seconds — the old 5s budget let runner load truncate
+            // the loop mid-sequence, producing the documented "AttemptCount not > 3" CI flake.
+            using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
             var result = await cm.Connect(cts.Token);
 
             // Assert: more than 3 attempts must have been consumed.
             // If rejection counter weren't being reset by failures, it would stop at 3.
+            cts.Token.IsCancellationRequested.ShouldBeFalse(
+                "The loop must stop via sequence exhaustion, not the hang-guard CTS");
             result.ShouldBeFalse("Connection should eventually fail");
             cm.AttemptCount.ShouldBeGreaterThan(3,
                 "More than MaxConsecutiveRejections attempts should run — counter was reset by interleaved failures");
--- a/McpPlugin.Tests/Network/Connection/ConnectionManagerTests.cs
+++ b/McpPlugin.Tests/Network/Connection/ConnectionManagerTests.cs
@@ -16,6 +16,7 @@ using Microsoft.AspNetCore.Http.Connections;
 using Microsoft.AspNetCore.SignalR.Client;
 using Microsoft.Extensions.Logging;
 using Moq;
+using R3;
 using Xunit;
 using Xunit.Abstractions;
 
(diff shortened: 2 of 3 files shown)


IvanMurzak:
## Review — depth `high` (adversarial), independent reviewer

Reviewed head `632c0ca` against `main` (`ed98817`). Walked: lock-order across Connect/Disconnect/DisconnectImmediate/Dispose/DisposeAsync, every completion path of the TCS proxy, the `ReferenceEquals` ownership guard, joiner semantics, consumer-visible deltas, both test changes (including tracing the pin test on the *old* code path), and the merge commit.

**BLOCKING: none.**

### Verified correct

- **Lock order.** No path acquires `_gate` under `_ongoingConnectionGate` or holds `_ongoingConnectionGate` across an await. `Disconnect` = `_gate` → gate (Disconnect.cs:49 → :204); `DisconnectImmediate` uses `Wait(TimeSpan.Zero)` (non-blocking, Disconnect.cs:138); `Dispose`/`DisposeAsync` take only `_gate`, bounded. The comment's deadlock rationale for publishing a proxy instead of the real task is accurate.
- **Proxy lifecycle.** Between install-release and the outer `try` only non-throwing statements run, so a leader always reaches the cleanup `finally`; the proxy is completed on success, failure, ConnectCore exception, DisconnectImmediate-detach, and disposed-gate ODE — always *after* the clear-if-owned, so late arrivals start fresh instead of adopting stale results. No ABA on `ReferenceEquals` (each TCS is a fresh object; nothing re-installs an old task). Two leaders can coexist only after Disconnect detaches the first — intended fresh-attempt semantics, still serialized on `_gate`, and the ownership guard correctly prevents the old leader clobbering the successor (required *because* publication moved out from under `_gate`; the old unconditional clear would now be wrong, and it was replaced).
- **Semantics.** Joiners' own tokens are honored in `WaitForConnectionCompletion` (false on cancel; doesn't cancel the leader — parity with old code). Joiners observe the leader's result; coalescing is the contract three pre-existing tests already assert. No observable event-ordering/KeepConnected change; `_ongoingConnectionTask` is private.
- **Pin test.** Genuinely deterministic — all coordination is TCS-driven on the test thread's synchronous prefix (R3 fires the KeepConnected subscriber synchronously inside `_continueToReconnect.Value = true`, which on old code is exactly the pre-publication window; the queued caller wakes after the leader clears the field under `_gate` → duplicate `CreateConnectionAsync` → `providerCallCount == 2`, the CI shape). No sleeps on the assertion path; the 30 s CTSes are pure hang guards.
- **Rejection tests.** Fixture confirms the timing math (RejectionThreshold 50 ms, `WaitBeforeRetry` no-op ⇒ ~1.15 s real time vs the old 5 s budget). The new `IsCancellationRequested.ShouldBeFalse` assertions sit *before* the AttemptCount asserts, so a future truncation reports its true cause instead of the misleading flake shape — a real assertion gain, matching the file's existing pattern. No assertion weakened.
- **Merge commit `632c0ca`.** Diff vs parent2 (`ed98817`) is exactly the 3 PR files (+166/−37); diff vs parent1 is exactly #204's 6 auth files (+592/0). Disjoint file sets, no evil-merge content. CI green on ubuntu/macos/windows.

### ADVISORY

1. **Orphanable proxy on a dispose race (hardening, not a merge-blocker).** `Connect`'s cleanup does `await _ongoingConnectionGate.WaitAsync(CancellationToken.None)` (ConnectionManager.Connect.cs:85). `SemaphoreSlim.Dispose()` does **not** release pending async waiters, and a subsequent `Release()` by the momentary holder throws ODE without waking them — so if `Dispose()`/`DisposeAsync()` (ConnectionManager.cs:216/:265) disposes the gate while this wait is *pending* (gate momentarily held by another caller's install/join section), the cleanup strands forever and the `TrySetResult` at :101 never runs. The `catch (ObjectDisposedException)` at :92 covers only WaitAsync-called-*after*-dispose, not pending-*across*-dispose. Net-new exposure: joiners now depend on the leader's cleanup completing (old joiners awaited the already-completed real task), and a joiner that passed the default `CancellationToken.None` would hang forever. Reachability is tiny — requires dispose concurrent with in-flight Connect traffic *and* the gate contended in a ~sub-microsecond window; the in-product joiner call sites all carry cancelable tokens (the closed-event reconnect's token is canceled by `DisposeCommonSync` before the gates are disposed). Suggested hardening for a follow-up: a bounded wait (e.g. `WaitAsync(TimeSpan.FromSeconds(5))`, best-effort clear on timeout) — keeps the complete-after-clear ordering on the common path while making the proxy's completion unconditional.
2. **Nit:** `ConnectCore`'s log lines still say `nameof(Connect)` (ConnectionManager.Connect.cs:116/:121/:126/:166). Fine if intentional (log-grep continuity with pre-fix logs); noting it so it's a choice, not an accident.

**Verdict: PASS** — no blocking findings at depth `high`.

PR #206, opened 2026-08-15
fix(ci): eliminate all warnings surfaced by GitHub Actions runs

IvanMurzak:
Every job of the reference run (https://github.com/IvanMurzak/MCP-Plugin-dotnet/actions/runs/31879231425) was green but carried warning annotations. This PR eliminates every warning the workflows surface, behavior-neutrally. Full inventory (enumerated from a clean local mirror of the CI steps AND from the run's complete logs — all 40 `##[warning]` markers across the run reduce to these root causes):

### 1. CS0109 ×3 — `McpPlugin.Common/src/Data/Response/Tool/Call/ResponseCallValueTool.Static.cs:19,24,61`
`Error(Exception)`, `Error(string?)`, `ErrorStructured(JsonNode?)` declared `new`, but the base `ResponseCallTool` overloads take extra optional parameters (`ResponseErrorKind`, `int?`) — different signatures, nothing hidden, so the compiler flags the modifier as spurious. **Fix:** removed `new` on those three. `new` is source-level only (no effect on member lookup, overload resolution, or IL), so public API and behavior are unchanged. `Success`/`SuccessStructured`/`Processing` hide identical base signatures and correctly keep `new`.

### 2. CS4014 ×2 TFMs — `McpPlugin.Tests/Network/Connection/ConnectionManagerTests.cs:442`
Shouldly's `ShouldNotBeNull` **returns** the `Task<bool>` it validated; the bare statement discarded an awaitable inside an async method. The task is *deliberately* un-awaited there — the single-flight race-pin test awaits it later, after the leader completes. **Fix:** explicit discard (`_ = …`) + comment. Execution order and test semantics untouched.

### 3. `WARNING: Overwriting results file` ×3 per test job (workflow logs)
`--logger "trx;LogFileName=test-results.trx"` makes all 4 test runs (2 projects × 2 TFMs) write the same file — 3 of 4 .trx result files were destroyed before upload/publish. **Fix:** `LogFilePrefix=test-results` in `test-pull-request.yml` and `release.yml` — one uniquely-named .trx per run, all still matching the existing `TestResults/**/*.trx` globs. No job/check names changed. (`deploy.yml`'s test step uses no trx logger; its three `dotnet pack` steps were verified warning-free locally.)

### Not actionable (reported for completeness)
- git's `init.defaultBranch` hint during `actions/checkout` — runner-side git chatter, not repo-fixable.
- `warn:`-level application log lines during tests — intentional product logging that tests (e.g. `RouterLogCaptureTests`) assert on; not CI warnings.

### Proof (local, mirroring CI exactly: clean tree → restore → build Release → test; plus the golden-vector job's commands)
- `dotnet build --no-restore --configuration Release`: **0 Warning(s), 0 Error(s)**
- `dotnet build McpPlugin.Tests/... --no-restore --configuration Release`: **0 Warning(s), 0 Error(s)**
- Full suite green both TFMs: McpPlugin.Tests **880/880** (net8.0 + net9.0), McpPlugin.Server.Tests **684/684** (net8.0 + net9.0), golden vectors **71/71** both TFMs
- Test run with the new logger form: **0** overwrite warnings; 4 uniquely-named .trx files produced

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,11 @@ jobs:
 
       - name: Test
         working-directory: ./
-        run: dotnet test --no-build --verbosity normal --configuration Release --logger "trx;LogFileName=test-results.trx" --results-directory ./TestResults
+        # LogFilePrefix (not LogFileName): 2 test projects x 2 TFMs = 4 test runs; a fixed
+        # LogFileName makes each run overwrite the previous .trx ("WARNING: Overwriting results
+        # file") so only the last leg's results survived. One uniquely-named .trx per run instead;
+        # all still match the TestResults/**/*.trx globs below.
+        run: dotnet test --no-build --verbosity normal --configuration Release --logger "trx;LogFilePrefix=test-results" --results-directory ./TestResults
 
       - name: Publish Test Results
         uses: EnricoMi/publish-unit-test-result-action@v2
--- a/.github/workflows/test-pull-request.yml
+++ b/.github/workflows/test-pull-request.yml
@@ -40,7 +40,11 @@ jobs:
 
       - name: Test
         working-directory: ./
-        run: dotnet test --no-build --verbosity normal --configuration Release --logger "trx;LogFileName=test-results.trx" --results-directory ./TestResults
+        # LogFilePrefix (not LogFileName): the solution has 2 test projects x 2 TFMs = 4 test runs;
+        # a fixed LogFileName makes each run overwrite the previous .trx ("WARNING: Overwriting
+        # results file") so only the last leg's results survived to upload/publish. The prefix form
+        # yields one uniquely-named .trx per run, all still matching TestResults/**/*.trx below.
+        run: dotnet test --no-build --verbosity normal --configuration Release --logger "trx;LogFilePrefix=test-results" --results-directory ./TestResults
 
       - name: Publish Test Results
         # The docker-based action only runs on Linux; the Windows/macOS legs upload their .trx
--- a/McpPlugin.Common/src/Data/Response/Tool/Call/ResponseCallValueTool.Static.cs
+++ b/McpPlugin.Common/src/Data/Response/Tool/Call/ResponseCallValueTool.Static.cs
@@ -16,12 +16,12 @@ namespace com.IvanMurzak.McpPlugin.Common.Model
 {
     public partial class ResponseCallValueTool<T> : ResponseCallTool
     {
-        public static new ResponseCallValueTool<T> Error(Exception exception)
+        public static ResponseCallValueTool<T> Error(Exception exception)
         {
             return Error($"[Error] {exception?.Message}\n{exception?.StackTrace}");
         }
 
-        public static new ResponseCallValueTool<T> Error(string? message = null)
+        public static ResponseCallValueTool<T> Error(string? message = null)
         {
             return new ResponseCallValueTool<T>(
                 status: ResponseStatus.Error,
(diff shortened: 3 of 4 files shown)


IvanMurzak:
## Review — depth: high (blocking) — verdict: PASS

Independent review of `210fb05c` (not the implementer). Verified against a read-only checkout of the exact PR head plus the live CI run. **No blocking findings. No advisory findings. Clean.**

### 1. `ResponseCallValueTool.Static.cs` — spurious `new` removal: correct, provably no behavioral/API change
- Base `ResponseCallTool` statics compared member-by-member: the three members that **dropped** `new` — `Error(Exception)`, `Error(string?)`, `ErrorStructured(JsonNode?)` — face base overloads with extra `ResponseErrorKind`/`int?` parameters (different arity ⇒ no hiding ⇒ CS0109 was correct). The three siblings that **keep** `new` — `Success(string?)`, `SuccessStructured(JsonNode?)`, `Processing(string?)` — hide *identical* base signatures, so their `new` is genuinely load-bearing. Base chain ends at `ResponseCallTool : IRequestID`, no further inherited statics.
- Language-level: `new` on a member is a pure warning-suppression modifier — it emits nothing into IL/metadata and plays no part in member lookup or overload resolution, so removing it cannot alter behavior or the NuGet API surface.
- Machine-checked in **both** directions by a full non-incremental Release rebuild at the PR head: **0 warnings** ⇒ no CS0108 (nothing load-bearing was removed) and no CS0109 (the kept `new`s are genuine).

### 2. `ConnectionManagerTests.cs:444` — explicit discard: zero semantic drift
- `_ = expr;` evaluates the identical expression in the identical order as the bare statement; the only delta is suppressing CS4014. The assertion still executes; the returned `Task<bool>` is (as the new comment says) deliberately awaited later — `Task.WhenAny(secondConnect!, …)` at L459 and `await secondConnect!` at L463 are untouched.
- Ran `Connect_CallerInPrePublicationWindow_JoinsInFlightAttempt` at the PR head on **both TFMs**: passes in ~0.4s each — i.e. driven by the explicit signals, not the 30s hang-guard CTSes, so the interleaving choreography is intact.

### 3. Workflows — `LogFileName` → `LogFilePrefix`: valid, globs match, names untouched, release flow unaffected
- Diff is one `run:` line + explanatory comment per file. Job names (`test`, `golden-vector-parity`), matrix, step names, action versions, and `permissions` blocks are byte-unchanged — merge gating keys are intact. No new actions.
- Syntax verified empirically on the SDK family CI uses (8.0.x/9.0.x): a probe run produced `test-results_net8.0_<timestamp>.trx` / `test-results_net9.0_<timestamp>.trx` **directly in the results-directory root**, so `TestResults/**/*.trx` matches (`**` matches zero segments in both EnricoMi's Python glob and `@actions/glob`).
- Real CI run 31882703339 confirms: **4 distinct .trx per OS** (2 projects × 2 TFMs), all under `TestResults/`. The aggregate `Test Results` check now reports **“All 1 564 tests pass”** vs **“All 880 tests pass”** on pre-fix `main` (run 31879727445) — the last-survivor overwrite defect is really fixed, and the fix is why the number moved.
- `release.yml`: artifact names unchanged (`test-results`); grep across `.github/` finds **nothing** that downloads these artifacts or references `test-results.trx` by exact filename; `release-at-github`/`deploy` consume only version outputs. The trx renaming cannot reach the release artifact flow.

### 4. Completeness spot-check
- Independent full `--no-incremental` Release rebuild of the whole solution at `210fb05c`: **0 Warning(s), 0 Error(s)** (verified a real recompile via output timestamps, not an up-to-date no-op).
- All 5 check runs on the head SHA: success, **0 annotations** each — the worker's claim independently reproduced.

**Verdict: PASS.**

PR #207, opened 2026-08-15
b3: shared HttpTokenRefresher, family-aware ITokenRefresher API, RFC 8693 exchange + RFC 7009 revocation clients, lock-integrated provider refresh

IvanMurzak:
unified-machine-auth **b3-cs-refresher** — the ONE shared C# refresh implementation for all three engine plugins (design 04 §3), the RFC 8693 exchange client (04 §4, a5-frozen shape), the F1 two-lock-hold login-commit helper, the F6 machine-wide sign-out helper, and the lock-integrated double-checked refresh path in `PluginCredentialProvider`.

## What changed

- **`ITokenRefresher` (published NuGet public API)** — new family-aware `RefreshAsync(TokenRefreshRequest, CancellationToken)` overload added as a **default interface method**; the legacy two-string method stays abstract and unchanged. `TokenRefreshResult` gains `FailureKind` (`InvalidGrant` vs `Transient`) via a new `Failure(reason, kind)` overload.
- **`HttpTokenRefresher` (new, shared)** — presents the family's **stored `clientId`** (component default only for a legacy family of unknown id, 04 §3.7); **omits `scope` and `resource` entirely** (04 §3.3 / P0-3 — the server falls back to the stored grant); success without `refresh_token` ⇒ caller keeps the previous one (04 §3.4); **one network attempt per family per skew window per process** (04 §3.6, keyed by refresh token, in-flight attempt shared); **15 s HTTP timeout** (= `REFRESH_HTTP_TIMEOUT`, 04 §2); `invalid_grant` classified distinctly (04 §3.5).
- **`TokenExchangeClient` (new)** — RFC 8693 agent→plugin derivation, exactly the a5-frozen wire shape (`grant_type`/`subject_token`/`subject_token_type` URNs, `client_id` ≤64, `scope=mcp:plugin`); `audience`/`resource` deliberately omitted. **`TokenRevocationClient` (new)** — RFC 7009, fail-soft.
- **`PluginCredentialProvider`** — refresh (proactive 60 s skew + reactive 401) now runs under the cross-process `MachineCredentialLock` with **re-read-first double-checking** (peer rotations adopted without a network call); `Unreadable` mid-refresh = **busy/retry**, never signed-out, never overwrite (b1 review A3); `IsStillOwned()` checked immediately before the store write (b2 review A1); `invalid_grant` → post-failure re-read → **dead-family verdict** with ONE structured telemetry event, other families never deleted, never loops (03 F3.5); family-aware read-modify-`Write()` — **never `Store.Rotate()`**; legacy §3.7 rewrite to v2 (+mirror) on first successful refresh; `SignOut()` routes through the 04 §2 lock delete path; boot on an unreadable store now surfaces `SignInRequired` (not `SignedOut`).
- **`MachineCredentialLoginCommit` (new)** — the F1 sequence: agent family (first hold) → exchange (no lock) → plugin family + v1 mirror (second hold); failed exchange leaves the committed agent family (`AgentOnly`); F7 subject guard refuses silent cross-account overwrite. **`MachineWideSignOut` (new)** — F6: revoke every family (stored `clientId`s; component default for legacy) with bounded retries, then the lock-protocol delete; offline sign-out still deletes locally.

## ⚠ Cascade note — public NuGet API change (for d1 Unity / e1 Godot / f1 Unreal)

**Compat strategy: default-interface-method overload — chosen over an in-place signature change** so all three engine repos compile unmodified against this release (verified below). Consequences:

1. **Nothing breaks at compile time.** The legacy `RefreshAsync(string, string?, CancellationToken)` is untouched and still abstract; `UnityTokenRefresher`, `GodotTokenRefresher`, and the Unreal bridge `OAuthTokenRefresher` keep compiling as-is.
2. **`PluginCredentialProvider` now calls the NEW overload.** For an un-migrated engine refresher, the DIM delegates to the legacy method — i.e. status-quo behavior (component-default `client_id`; Unity additionally still sends `scope`, the P0-3 defect). **b3 makes the cascade non-breaking; it does not deliver the engine-side fixes — d1/e1/f1 must still replace the engine refreshers with `new HttpTokenRefresher(defaultServerTarget, componentClientId)`** (delete or thin-adapter the local ones).
3. **Test-suite gotcha (hit in this repo, will hit engines):** Moq/Castle **intercepts the DIM**, so a loose `Mock<ITokenRefresher>` returns a null `Task` to the provider unless you add `Setup(r => r.RefreshAsync(It.IsAny<TokenRefreshRequest>(), It.IsAny<CancellationToken>()))`. Engine test suites that mock `ITokenRefresher` need the same one-line setup.
4. **`PluginCredentialProvider` ctor** gained trailing optional parameters (`credentialLock`) — source-compatible, binary-breaking (engines recompile against the package, so this is automatic; no call-site change needed unless injecting a custom lock).
5. **Behavior engines inherit by recompiling:** refresh under the machine-wide file lock with double-checked re-read; lock-busy = transient false (no `SignInRequired`); unreadable-store boot = `SignInRequired`; `SignOut()` via the lock delete path.
6. **New helpers to adopt in the cascades:** F1 login → `MachineCredentialLoginCommit.CommitAsync(store, lock, agentFamily, serverTarget, subject, new TokenExchangeClient(target, ownClientId))`; F6 sign-out → `MachineWideSignOut.SignOutAsync(store, lock, new TokenRevocationClient(target), componentClientId)`.
7. **TFM/profile:** all new API compiles on `netstandard2.1` + `LangVersion 9` (Unity 2022.3+ supports default interface methods; verified by the Unity-profile smoke below).

## Evidence

- `dotnet test McpPlugin.Tests` — **933/933 green on net8.0 AND net9.0** (66 new tests across `HttpTokenRefresherTests`, `TokenExchangeClientTests`, `PluginCredentialProviderFamilyRefreshTests`, `MachineCredentialLoginCommitTests`, `MachineWideSignOutTests`, plus the b1-A1 `Adopt`/`Write` mutation-seam pin). `McpPlugin.Server.Tests` — 684/684 green both TFMs. Full solution builds (netstandard2.1 leg included).
- **Mandated plants (G-SEC-1), RED-verified then reverted:**
  - Plant A (`client_id` ← component default, ignoring the stored id): `Refresh_PresentsTheFamilysStoredClientId_NeverTheComponentDefault` **RED** (1 failed / 18 passed) — the cross-mint discriminator, mirrors probe Q1.
  - Plant B (`scope=mcp:plugin` added to the refresh form): `Refresh_OmitsScopeAndResourceEntirely` **and** `Refresh_LegacyTwoStringApi_PresentsTheComponentDefault_AndStillOmitsScope` **RED** (2 failed / 17 passed) — the P0-3 discriminator.
- **Engine smoke-compiles (scratch copies, engines untouched):** Godot-MCP full plugin project against the locally-packed `com.IvanMurzak.McpPlugin 99.0.0-b3smoke` — **Build succeeded**; Unreal bridge sidecar via its own `-p:UseLocalMcpPlugin=true` source-reference mechanism against this branch — **Build succeeded**; Unity's real `UnityTokenRefresher.cs` + `DeviceAuthService.cs` compiled verbatim under the Unity profile (`netstandard2.1`, C# 9) against the packed package — **Build succeeded**.

Release/versioning is task b4 — this PR intentionally does not touch `<Version>`.

### DIFF
--- a/McpPlugin.Tests/Infrastructure/RecordingHttpMessageHandler.cs
+++ b/McpPlugin.Tests/Infrastructure/RecordingHttpMessageHandler.cs
@@ -1,0 +1,160 @@
+/*
+┌────────────────────────────────────────────────────────────────────────┐
+│  Author: Ivan Murzak (https://github.com/IvanMurzak)                   │
+│  Repository: GitHub (https://github.com/IvanMurzak/MCP-Plugin-dotnet)  │
+│  Copyright (c) 2025 Ivan Murzak                                        │
+│  Licensed under the Apache License, Version 2.0.                       │
+│  See the LICENSE file in the project root for more information.        │
+└────────────────────────────────────────────────────────────────────────┘
+*/
+
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace com.IvanMurzak.McpPlugin.Tests.Infrastructure
+{
+    /// <summary>
+    /// An <see cref="HttpMessageHandler"/> that records every request (URL + decoded form fields)
+    /// and replays scripted responses — the wire-assertion seam for the unified-machine-auth b3
+    /// HTTP clients (<c>HttpTokenRefresher</c>, <c>TokenExchangeClient</c>,
+    /// <c>TokenRevocationClient</c>). No sockets, no ports (workspace rule: workers bind only
+    /// assigned ports — this binds none).
+    /// </summary>
+    public sealed class RecordingHttpMessageHandler : HttpMessageHandler
+    {
+        readonly Queue<Func<RecordedRequest, HttpResponseMessage>> _scripted = new Queue<Func<RecordedRequest, HttpResponseMessage>>();
+        readonly List<RecordedRequest> _requests = new List<RecordedRequest>();
+        Func<RecordedRequest, HttpResponseMessage>? _fallback;
+        TimeSpan _delay = TimeSpan.Zero;
+        Action<RecordedRequest>? _onRequest;
+
+        /// <summary>One recorded request: absolute URL and the decoded form body.</summary>
+        public sealed class RecordedRequest
+        {
+            public RecordedRequest(string url, IReadOnlyDictionary<string, string> form)
+            {
+                Url = url;
+                Form = form;
+            }
+
+            public string Url { get; }
+            public IReadOnlyDictionary<string, string> Form { get; }
+        }
+
+        /// <summary>Every request this handler served, in order.</summary>
+        public IReadOnlyList<RecordedRequest> Requests => _requests;
+
+        /// <summary>Enqueue one scripted response (consumed in FIFO order before <see cref="RespondWith"/>).</summary>
+        public RecordingHttpMessageHandler Enqueue(HttpStatusCode status, string json)
+        {
+            _scripted.Enqueue(_ => JsonResponse(status, json));
+            return this;
+        }
+
+        /// <summary>The response used when the scripted queue is empty.</summary>
+        public RecordingHttpMessageHandler RespondWith(HttpStatusCode status, string json)
+        {
+            _fallback = _ => JsonResponse(status, json);
+            return this;
+        }
+
+        /// <summary>Delay every response (for timeout tests).</summary>
+        public RecordingHttpMessageHandler DelayEachResponse(TimeSpan delay)
+        {
+            _delay = delay;
+            return this;
+        }
+
+        /// <summary>Invoked synchronously as each request is recorded (e.g. to displace a lock file mid-call).</summary>
+        public RecordingHttpMessageHandler OnRequest(Action<RecordedRequest> callback)
+        {
+            _onRequest = callback;
+            return this;
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            var body = request.Content == null ? "" : await request.Content.ReadAsStringAsync().ConfigureAwait(false);
+            var recorded = new RecordedRequest(request.RequestUri?.ToString() ?? "", DecodeForm(body));
+            lock (_requests)
+                _requests.Add(recorded);
+            _onRequest?.Invoke(recorded);
+
+            if (_delay > TimeSpan.Zero)
+                await Task.Delay(_delay, cancellationToken).ConfigureAwait(false);
+
+            Func<RecordedRequest, HttpResponseMessage>? producer;
+            lock (_requests)
+                producer = _scripted.Count > 0 ? _scripted.Dequeue() : _fallback;
+            if (producer == null)
+                throw new InvalidOperationException("RecordingHttpMessageHandler: no scripted response left and no fallback configured.");
+            return producer(recorded);
+        }
+
+        static HttpResponseMessage JsonResponse(HttpStatusCode status, string json)
+            => new HttpResponseMessage(status)
+            {
+                Content = new StringContent(json, Encoding.UTF8, "application/json"),
+            };
+
+        static IReadOnlyDictionary<string, string> DecodeForm(string body)
+        {
+            var result = new Dictionary<string, string>(StringComparer.Ordinal);
+            foreach (var pair in body.Split(new[] { '&' }, StringSplitOptions.RemoveEmptyEntries))
+            {
+                var eq = pair.IndexOf('=');
+                var key = eq < 0 ? pair : pair.Substring(0, eq);
+                var value = eq < 0 ? "" : pair.Substring(eq + 1);
+                result[Uri.UnescapeDataString(key.Replace('+', ' '))] = Uri.UnescapeDataString(value.Replace('+', ' '));
+            }
+            return result;
+        }
+    }
+
+    /// <summary>
+    /// A deterministic scripted <see cref="ITokenRefresher"/> for provider tests. Hand-rolled (not
+    /// Moq) on purpose: the interface now carries a default implementation for the request-based
+    /// overload, and a hand-written fake pins EXACTLY which overload the provider invokes —
+    /// proxy-generated mocks leave that dispatch ambiguous.
+    /// </summary>
+    public sealed class FakeTokenRefresher : ITokenRefresher
+    {
+        readonly Func<TokenRefreshRequest, TokenRefreshResult> _onRefresh;
+        readonly List<TokenRefreshRequest> _requests = new List<TokenRefreshRequest>();
+
+        public FakeTokenRefresher(Func<TokenRefreshRequest, TokenRefreshResult> onRefresh)
+        {
+            _onRefresh = onRefresh;
+        }
+
+        public FakeTokenRefresher(TokenRefreshResult fixedResult)
+            : this(_ => fixedResult)
+        {
+        }
+
+        /// <summary>Every request-based refresh call the provider made, in order.</summary>
+        public IReadOnlyList<TokenRefreshRequest> Requests => _requests;
+
+        /// <summary>How many times the LEGACY two-string API was called (must stay 0 — the provider uses the family-aware API).</summary>
+        public int LegacyCalls { get; private set; }
+
+        public Task<TokenRefreshResult> RefreshAsync(string refreshToken, string? serverTarget, CancellationToken cancellationToken = default)
+        {
+            LegacyCalls++;
+            return Task.FromResult(_onRefresh(new TokenRefreshRequest(refreshToken, serverTarget, clientId: null)));
+        }
+
+        public Task<TokenRefreshResult> RefreshAsync(TokenRefreshRequest request, CancellationToken cancellationToken = default)
+        {
+            _requests.Add(request);
+            return Task.FromResult(_onRefresh(request));
+        }
+    }
+}
--- a/McpPlugin.Tests/Network/Connection/Credentials/ConnectionCredentialCoordinatorTests.cs
+++ b/McpPlugin.Tests/Network/Connection/Credentials/ConnectionCredentialCoordinatorTests.cs
@@ -94,7 +94,9 @@ namespace com.IvanMurzak.McpPlugin.Tests.Network.Connection.Credentials
         [Fact]
         public async Task HandleRejection_RefreshFails_DoesNotReconnect_SurfacesSignInRequired()
         {
-            var refresher = RefresherReturning(TokenRefreshResult.Failure("refresh token expired"));
+            // "refresh token expired" IS an invalid_grant — the dead-family kind, the ONLY kind
+            // that reaches SignInRequired (review B2; a transient failure stays signed in).
+            var refresher = RefresherReturning(TokenRefreshResult.Failure("refresh token expired", TokenRefreshFailureKind.InvalidGrant));
             using var provider = SeededProvider(refresher.Object);
             using var connection = new FakeConnection();
             using var coordinator = new ConnectionCredentialCoordinator(connection, provider);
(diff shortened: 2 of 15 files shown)


IvanMurzak:
## Taskflow review — depth **high** (blocking) — b3-cs-refresher @ `ea68b78`

Reviewed against the immutable spec (design `04-credential-store-and-refresh.md` §3/§3.7/§4, `03-auth-flows.md` F1/F3.5/F6/F7, `06-migration-rollout.md` §Ordering) and the b1/b2 carry-forwards. Verified independently: refresh wire form is exactly `{grant_type, refresh_token, client_id}` with a full key-set pin on a **real** decoded `FormUrlEncodedContent` body (the `RecordingHttpMessageHandler` decodes the actual HttpClient pipeline output — not mock-echo, so the wire pins are non-vacuous); exchange form is exactly the 5 a5 params with byte-exact URNs and a ctor-enforced ≤64 `client_id`; revoke shape matches AGS `routers/oauth.py:528` ground truth (form-encoded `token` (+ harmless `token_type_hint`/`client_id`), no auth, 200-uniform, fail-soft); no `Store.Rotate()` anywhere; `Unreadable` mid-refresh = busy with byte-level file-preservation test; `IsStillOwned()` before every store write; family rate-key = refresh token (family-unique; window self-resets on rotation because the key rotates); plant counts (1 failed/18 passed, 2 failed/17 passed) are consistent with the 19-case refresher suite; zero overlap with PR #206's four files; no store/lock-internal touches (the plaintext test seams pre-exist at base).

### Ruling on the DIM compat strategy (the cascade-note question) — **sound, PASS**

- **Legal on this TFM set:** `netstandard2.1` declares `RuntimeFeature.DefaultImplementationsOfInterfaces`; a DIM that were illegal would fail the ns2.1 leg with CS8701, and CI is green on all three OSes building that leg. Unity 2022.3 (ns2.1 profile, C# 9) supports DIMs at runtime.
- **Delegation direction is the only non-breaking one.** The legacy two-string method stays abstract (engines keep compiling); the new overload's default delegates *down*, deliberately dropping `request.ClientId`. For an un-migrated engine refresher invoked through the new seam this is **exactly status quo** (component-default id; Unity's scope defect persists *in its own refresher only*) — the 06 O4 invariant ("no cohort below status quo"), not a silent P0-3 violation: a legacy implementation has no parameter that could receive the stored id, and the alternative (throwing DIM) would runtime-break unadopted engines. The regression risk is confined to a consumer that bumps the NuGet without doing its d1/e1/f1 row — a state the 06 ordering excludes, and the PR body documents the residue accurately.
- The Moq-intercepts-the-DIM gotcha is real, correctly diagnosed, and mitigated in `ConnectionCredentialCoordinatorTests` + the cascade note.

### BLOCKING

**B1. `MachineCredentialLoginCommit.CommitPluginFamilyAsync` has no F7 subject guard at the second hold** (`MachineCredentialLoginCommit.cs:206-233`). The first hold guards subject mismatch; the second hold re-reads the store but then writes `families.Plugin` into **whatever document is now there**, preserving its `Subject` (`:224-225` only fills a *missing* subject). Interleavings that break it: (a) another surface completes a login as account B between the holds (exchange takes up to 15 s); (b) the **retryable** `PluginCommitBusy` recovery — the enum's own doc invites calling this method arbitrarily later; (c) the tools-only entry. Outcome: store says `subject: B` with agent family B but `families.plugin` derived from account A — the machine's tools authenticate as A while every subject-keyed check (F5 pickup, F7 guard, telemetry) believes B. That is precisely the "silent cross-account overwrite" 03 F7.2 assigns to *every* login-surface write ("Before writing, the login surface compares new subject vs F's subject"), and this helper owns that guard (it defines `SubjectMismatch`). Fix: thread the expected subject into phase 2 (parameter, falling back to `exchange.Subject`), compare against the re-read document's `Subject` under the hold, return `SubjectMismatch` and write nothing on mismatch; add an interleaving test (login-B-between-holds ⇒ phase 2 refuses).

**B2. Transient refresh failures surface `SignInRequired` + fire `OnSignInRequired`** (`PluginCredentialProvider.cs:415`, the fall-through of the new `HandleRefreshFailure`). This contradicts, in order: (a) this PR's own `TokenRefreshFailureKind` contract (`ITokenRefresher.cs:19-38`: *"InvalidGrant is the ONLY kind that may declare a credential family dead"*, Transient = *"the caller keeps the existing credential and may try again later"*); (b) 03 F3, where "Sign in required" is exclusively the F3.5 dead-family outcome and busy-like failures "fail this attempt … retry next cycle"; (c) 03 F9 — offline "self-heals silently" on reconnect. Concretely: an AS outage, a 5xx, or one 15 s timeout during a proactive refresh flips the engine UI to "Sign in required" and fires the sign-in event on every attempt — in the ONE shared implementation all three cascades will bind UI to, so it ships fleet-wide with W1/W2. I am aware this is carried-over b7 behavior pinned by the pre-existing `RefreshAsync_Failure_SurfacesSignInRequired` (whose fixture is Transient-kind) — but b3 is the task that introduced the classification and rewrote this failure path; the conflation is now a one-line divergence from the enum's own documentation. Fix: transient ⇒ log + `return false` with **no state change** (exactly the lock-busy/unreadable treatment; also fold the `catch (Exception)` → `"refresh error"` branch at `:338-343` into it); keep `SignInRequired` for the invalid_grant dead verdict, no-refresh-token, and no-refresher cases; update the legacy test and add a pin that a Transient failure does NOT fire `OnSignInRequired`.

### ADVISORY (recorded, non-gating)

- **A1.** Rate discipline is per-`HttpTokenRefresher`-**instance**; 04 §3.6 says per **process** (the class doc says "per refresher instance", the PR body says "per process"). Holds only while each process constructs exactly one refresher — state that expectation explicitly in the d1/e1/f1 cascade note, or key the attempt map process-wide.
- **A2.** `PluginCredentialProvider.Adopt()` still writes the store lock-free while refresh and `SignOut()` are now lock-integrated. A peer inside its critical section can land its write after Adopt's blind write and silently revert a fresh in-editor sign-in until the next refresh cycle. Cascades should route login commits through `MachineCredentialLoginCommit`; either lock-guard Adopt or mark it a legacy seam in the cascade note.
- **A3.** Unity smoke-compile is partial (the at-risk files — `UnityTokenRefresher.cs`, `DeviceAuthService.cs` — verbatim under a ns2.1/C#9 profile against the packed package; not the full plugin project) — honestly flagged in the PR body. DoD met in substance (Godot full project, Unreal full bridge), but the caveat belongs on the d1 row so the full-project compile is confirmed at the top of the Unity cascade.

Everything else checked clean: keep-previous on missing `refresh_token`; §3.7 legacy rewrite to v2 (+mirror when sole plugin-plane credential, `ClientId` stays null); invalid_grant → post-failure re-read → one telemetry event (no token material, reset on adopt) → other families untouched on disk; F6 revoke-then-lock-delete with per-family stored ids, bounded retries, offline delete, unreadable-skip; F1 two-hold failure paths (`AgentOnly`, `Busy`, first-hold subject guard) all pinned by meaningful tests; no scope creep.

**Verdict: BLOCK** — two blocking findings (B1, B2); both are small, localized fixes well inside one fix round.

IvanMurzak:
## Taskflow review — fix round 1 re-review (depth **high**) — @ `2c7d571`

Re-reviewed the delta `ea68b78..2c7d571` (6 files, +488/−60, confined to the two fixed sources + their tests; CI green on all three OSes).

### B1 — no F7 guard at hold 2 — **CLOSED**

- **Every write site in the commit helper is now guarded.** Exactly two `store.Write` sites exist at the new head: hold 1 (`MachineCredentialLoginCommit.cs:211` — both-known subject compare, exact-premise check for confirmed switches, `IsStillOwned`) and hold 2 (`:370` — unreadable ⇒ retryable `StoreUnreadable`, missing ⇒ `StoreSignedOut` **never recreated** (this also fixes the old recreate-after-sign-out path, which I had missed calling out), subject-premise compare, `IsStillOwned`). The premise (`premiseSubject` = what the hold-1 write established) is threaded into phase 2, and `expectedSubject` is a required parameter on the public retry entry — callers cannot skip the guard silently.
- **The API change is sound and strictly *narrower*.** `overwriteDifferentSubject: bool` → `confirmedReplaceOfSubject: string?`: a confirmation now authorizes replacing exactly one named account, re-verified *inside* the hold (`GuardPremiseChanged` on drift — pinned by `Commit_ConfirmedSwitch_ButStoreOwnerChangedAgain_AbortsGuardPremiseChanged`). No widening; `null` keeps refuse-on-mismatch.
- **Orphan revocation is correctly scoped**: only the two TERMINAL hold-2 aborts (`SubjectMismatch`-at-hold-2, `StoreSignedOut`), outside the critical section, refresh-token preferred, optional client, single attempt; retryable outcomes (`PluginCommitBusy`/`StoreUnreadable`) and hold-1 aborts revoke nothing — each direction pinned (`Commit_HoldOneSubjectMismatch_RevokesNothing`, `CommitPluginFamily_UnreadableStore_…NeverRevokes`).
- The interleaving I demanded is pinned head-on: `Commit_SubjectChangedBetweenHolds_AbortsSecondPhase_NeverMixesAccounts_RevokesTheOrphan` asserts B's store untouched, A's plugin family absent, orphan revoked.
- **Plants C/D discriminate**: disabling orphan revoke reddens exactly the two revoke assertions; skipping the exact-premise check reddens the `GuardPremiseChanged` pin. RED record (6 guard + 2 transient = 8) and the test count (933 + 9 new commit tests + 1 net provider split = 943) are both arithmetically consistent with the suite.

### B2 — transient failures surfacing SignInRequired — **CLOSED**

- Both failure funnels fixed: the `catch (Exception)` branch (`PluginCredentialProvider.cs:342-348`) and the `HandleRefreshFailure` fall-through (`:418-427`) now log + `return false` with **no state change**. `SignInRequired`/`OnSignInRequired` are reachable only from: confirmed dead family (invalid_grant + post-failure re-read), no-refresh-token, no-refresher — exactly 03 F3.5. No path loops (single-attempt structure unchanged; refresher rate discipline untouched).
- **The rewritten tests are honest intent-preserving splits, not weakenings.** Old `RefreshAsync_Failure_SurfacesSignInRequired` (fixture was Transient-kind — the defect pin) → its protective intent lives on in `RefreshAsync_InvalidGrant_SurfacesSignInRequired`, and the corrected direction is pinned by `RefreshAsync_TransientFailure_StaysSignedIn_DoesNotSurfaceSignInRequired` (state + event + credential-kept, all three asserted). `ThrowingRefresher` flip is the deliberate B2 correction, with the no-event direction pinned. The coordinator fixture now honestly declares `InvalidGrant`, preserving its reject-path intent. The dead-family test gained the positive event assertion — both directions of the split are now pinned.

### Ruling on the flagged twin divergence (revoke-on-`StoreUnreadable`-abort): **acceptable-and-record**

TS revokes the orphan and re-exchanges on retry; C# keeps it and re-commits the same mint. Each is internally consistent — the dangerous quadrants would be *revoke+commit-same* (commits dead tokens) or *keep+re-exchange-forever* (orphan accumulation), and neither implementation is in one. The store is written in neither case, so a mixed C#/TS machine observes no difference; the only residue is server-side (an unretried C# abort leaves one family to expire naturally ≤30 d — the same bounded residue as F6.4 offline sign-out). No interop or parity-vector surface is touched. **Record it for x2 and the design ledger** as a deliberate strategy divergence; if convergence is ever wanted, the C# shape (fewer mint/revoke round-trips, no token churn) is the better target, i.e. TS would move — but nothing must move now.

### Advisory residue (non-gating, for the cascade note / ledger)

- A confirmed hold-1 switch retains the displaced account's *other* families in the document until overwritten (their revocation is the surface's F7.2 job — worth one line in the d1/e1/f1 cascade note so surfaces actually do it).
- The subject-less-premise window (`Commit_SubjectlessPremise_…ProceedsPerF73`) is the design-accepted F7.3 transitional hole; hold 2 could shrink it by falling back to `exchange.Subject` as the premise when `expectedSubject` is null (both-known-only, no false positives). Optional strengthening — record for x2.
- Prior advisories A1–A3 stand as recorded.

**Verdict: PASS** — both blocking findings resolved and pinned; no scope creep; no weakened assertions.

PR #208, opened 2026-08-15
test: RefreshHarness console harness for the x2 mixed-language concurrency suite

IvanMurzak:
C# half of the **x2 mixed-language concurrency suite** (taskflow `unified-machine-auth`, 04 §5 "Concurrency — real processes, not threads" / 03 F8). Coordinated PR pair with `AI-Game-Dev-CLI-Core#worktree-x2-concurrency-suite`, which holds the driver, the fake AS, and the suite's one CI home.

## What this adds

- **`McpPlugin.Tests.RefreshHarness/`** — a console exe (modeled on `McpPlugin.Tests.LockHarness`) exposing the REAL credential stack — `MachineCredentialStore`, `MachineCredentialLock`, `HttpTokenRefresher`, `PluginCredentialProvider` — as an OS worker process. The cli-core driver spawns it next to Node workers, all contending on ONE shared machine store against a local fake authorization server (line-cited mirror of `oauth_token_service.py` rotation + reuse⇒family-revoke + D10 grace window). Modes:
  - `hammer` — the real locked refresh loop (`GetAccessTokenAsync`), public ctors only, REAL 15/60/75 s contract constants;
  - `hammer-nolock` — the lock-DISABLED plant composition (same store + refresher, no mutual exclusion) for the mandated falsifiability plants;
  - `once [--scaled t,s,b]` — the lost-response plant's kill-victim and its retry; scaled constants go through the internal test ctors and the `timeout < stale < budget` ordering invariant is asserted before use (04 §2).
- **`MachineCredentialStore`: internal `plaintextForTesting` ctor overload** — the C# counterpart of the TS `identityCredentialCodec`. The mixed suite shares one store file between C# and Node processes, which needs a common at-rest format; the TS DPAPI codec shells PowerShell per operation (~1–2 s), which would make a ≥100-rotation hammer unrunnable. DPAPI↔DPAPI cross-implementation parity stays proven by the x1 suite. **No production behavior change**: the public constructor pins the platform codec exactly as before; the seam is `internal` + `InternalsVisibleTo`.
- Solution registration (fresh GUID, full config matrix) so the root `dotnet build` builds the harness on every CI leg.

## CI home (documented per the task's DoD)

The suite itself runs in **cli-core's `.github/workflows/concurrency-suite.yml`** — one CI home invoking both toolchains on 3 hosted OS legs (ubuntu / windows / macos), cloning this repo by ref (this branch, falling back to `main` after merge). This repo's own CI builds the harness (via the sln) and runs its full test suite unchanged.

## Evidence (local, Windows)

- Full solution test suite with the seam in place: `McpPlugin.Tests` 943 passed ×(net8+net9), `McpPlugin.Server.Tests` 684 passed ×2, 0 failed.
- Mixed suite 6/6 green via the cli-core driver: acceptance 111 rotations / **0 family revokes / 0 grace hits** at N=4 (2 C# + 2 TS); all three plants red-verified — incl. plant 3 both directions (a killed TS client's orphan lock taken over by this harness with scaled constants inside the D10 window, and vice versa, identical successor pair by value).

---

## Round 7 (resume): WARN stdout channel for the x2 grace-hit attribution

`1bd964e` — the harness previously passed `logger: null`, so a rotation received but **not persisted** (a Windows sharing-violation on the store write) was invisible from outside the process. A minimal `ILogger` now renders the provider's warnings as `WARN t=<unixMs> m=<base64 message>` stdout lines; the cli-core driver (PR IvanMurzak/AI-Game-Dev-CLI-Core#11) parses and classifies them (`Persisting … failed` / `lock was taken over` ⇒ LOST-WRITE; `refresh threw` / `failed transiently` ⇒ LOST-RESPONSE) to attribute every green-run D10 grace hit to a recorded client-side loss. No behavioral change to the credential stack — the C# store already implements the 04 §1 rename-retry contract and recorded **zero** losses under the suite's pressure runs; the root cause (the TS store missing that contractual retry) is fixed in the sibling PR.

### DIFF
--- a/McpPlugin.Tests.RefreshHarness/McpPlugin.Tests.RefreshHarness.csproj
+++ b/McpPlugin.Tests.RefreshHarness/McpPlugin.Tests.RefreshHarness.csproj
@@ -1,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!-- Real-subprocess harness for the mixed-language refresh concurrency suite
+       (unified-machine-auth 04 §5 "Concurrency - real processes, not threads", task x2).
+       Exposes the REAL McpPlugin credential stack - MachineCredentialStore,
+       MachineCredentialLock, HttpTokenRefresher, PluginCredentialProvider - as a console
+       process so the cross-language suite in AI-Game-Dev-CLI-Core
+       (test/concurrency-suite.subprocess.test.ts) can spawn C# workers next to Node workers
+       against one shared store and a local fake authorization server.
+       Test infrastructure only - never packed. -->
+
+  <PropertyGroup>
+    <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <LangVersion>11.0</LangVersion>
+    <IsPackable>false</IsPackable>
+    <RootNamespace>com.IvanMurzak.McpPlugin.Tests.RefreshHarness</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../McpPlugin/McpPlugin.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/McpPlugin.Tests.RefreshHarness/Program.cs
+++ b/McpPlugin.Tests.RefreshHarness/Program.cs
@@ -11,10 +11,12 @@
 #nullable enable
 using System;
 using System.IO;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using com.IvanMurzak.McpPlugin;
 using com.IvanMurzak.McpPlugin.AgentConfig;
+using Microsoft.Extensions.Logging;
 
 namespace com.IvanMurzak.McpPlugin.Tests.RefreshHarness
 {
@@ -68,6 +70,35 @@ namespace com.IvanMurzak.McpPlugin.Tests.RefreshHarness
         {
             Console.Out.WriteLine(line);
             Console.Out.Flush(); // the parent driver reacts to lines in real time — never buffer
+        }
+
+        /// <summary>
+        /// Minimal <see cref="ILogger"/> surfacing the provider's warnings as machine-parseable
+        /// stdout lines: <c>WARN t=&lt;unixMs&gt; m=&lt;base64 rendered message&gt;</c>. These are the
+        /// C# side's client-visible LOSS records — "Persisting refreshed credential failed" /
+        /// "Credential lock was taken over mid-refresh" mean a received rotation was NOT persisted
+        /// (the store still holds the AS-revoked predecessor); "Token refresh threw/failed
+        /// transiently" records a failed attempt that may have committed at the AS. The x2 driver
+        /// correlates them against the fake AS's grant-decision log to ATTRIBUTE every green-run
+        /// D10 grace hit; a grace hit no worker can account for is a serialization finding.
+        /// </summary>
+        private sealed class StdoutWarningLogger : ILogger
+        {
+            public static readonly StdoutWarningLogger Instance = new StdoutWarningLogger();
+
+            public IDisposable? BeginScope<TState>(TState state) where TState : notnull => null;
+
+            public bool IsEnabled(LogLevel logLevel) => logLevel >= LogLevel.Warning;
+
+            public void Log<TState>(
+                LogLevel logLevel, EventId eventId, TState state, Exception? exception,
+                Func<TState, Exception?, string> formatter)
+            {
+                if (!IsEnabled(logLevel))
+                    return;
+                var message = formatter(state, exception);
+                Report("WARN t=" + NowUnixMs + " m=" + Convert.ToBase64String(Encoding.UTF8.GetBytes(message)));
+            }
         }
 
         private static async Task<int> Main(string[] args)
@@ -170,7 +201,7 @@ namespace com.IvanMurzak.McpPlugin.Tests.RefreshHarness
                 using var provider = new PluginCredentialProvider(
                     store,
                     counting,
-                    logger: null,
+                    logger: StdoutWarningLogger.Instance,
                     refreshSkew: TimeSpan.FromMilliseconds(skewMs),
                     clock: null,
                     credentialLock: new MachineCredentialLock(baseDir));
@@ -292,7 +323,8 @@ namespace com.IvanMurzak.McpPlugin.Tests.RefreshHarness
 
             var counting = new CountingRefresher(httpRefresher);
             using var provider = new PluginCredentialProvider(
-                store, counting, logger: null, refreshSkew: null, clock: null, credentialLock: credentialLock);
+                store, counting, logger: StdoutWarningLogger.Instance,
+                refreshSkew: null, clock: null, credentialLock: credentialLock);
 
             Report("READY t=" + NowUnixMs + " pid=" + Environment.ProcessId + " mode=once" + (scaled != null ? "-scaled" : ""));
 
--- a/McpPlugin.Tests.RefreshHarness/Program.cs
+++ b/McpPlugin.Tests.RefreshHarness/Program.cs
@@ -1,0 +1,307 @@
+/*
+┌────────────────────────────────────────────────────────────────────────┐
+│  Author: Ivan Murzak (https://github.com/IvanMurzak)                   │
+│  Repository: GitHub (https://github.com/IvanMurzak/MCP-Plugin-dotnet)  │
+│  Copyright (c) 2025 Ivan Murzak                                        │
+│  Licensed under the Apache License, Version 2.0.                       │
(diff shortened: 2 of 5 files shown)


IvanMurzak:
## Taskflow review — depth high (blocking) — verdict: PASS

Reviewed jointly with `IvanMurzak/AI-Game-Dev-CLI-Core#11` (heads `1bd964e` / `bb15d1d`) as one task diff (x2-concurrency-suite, taskflow `2026-08-14-unified-machine-auth`). Full joint findings are posted on the cli-core PR; below is what applies to this repo.

### Blocking findings

None.

### Independent verification (this repo's half)

- **No production behavior change**: the only product-file diff is the `MachineCredentialStore` `plaintextForTesting` seam. All three instance DPAPI sites are gated symmetrically (`TryRead` :184, `ReadPlaintextJson` :315, `WriteProtected` :347 — verified against the full file at head; the static `ProtectBytes`/`UnprotectBytes` x1 seams are correctly untouched), the public constructor pins `plaintextForTesting: false`, and the seam is `internal` + `InternalsVisibleTo` for a never-packed harness assembly. With the flag false the code is byte-for-byte the old behavior.
- **Harness logger change is harness-only** (`RefreshHarness/Program.cs`): `StdoutWarningLogger` renders provider warnings as `WARN t=<unixMs> m=<base64>`; base64 framing survives newlines in exception messages, `Console.Out` is synchronized, and the driver trims `\r` — the stdout protocol (`READY`/`TOKEN`/`WARN`/`GRANT-DEAD`/`DONE`/`ONCE`) cannot be corrupted by message content. The four classified warning templates were verified verbatim against `PluginCredentialProvider.cs` at head (:347 "Token refresh threw", :370 "Credential lock was taken over mid-refresh", :374 "Persisting refreshed credential failed", :425 "Token refresh failed transiently") — the cli-core driver's regexes match exactly.
- **Contract parity confirmed**: `MachineCredentialStore.RenameRetryAttempts = 5` / `RenameRetryDelayMs = 250`, retrying `IOException`/`UnauthorizedAccessException` with the final attempt rethrowing — the reference the TS `renameWithRetrySync` fix mirrors. I independently reproduced the underlying Windows failure (single rename under a concurrent reader ⇒ `EPERM`) and the retry's survival on a real Windows machine.
- **Scaled-constants guard**: the `once --scaled` path asserts `timeout < stale < budget` before use (04 §2) — verified in `Program.cs` (`Once`, the `InvalidOperationException` guard).
- **CI**: all checks green at head `1bd964e` (3 OS test legs + golden-vector-parity); the suite's one CI home (cli-core `concurrency-suite.yml`) cloned this branch at exactly `1bd964e` on its green 3-leg run `31909910062`, so the mixed-suite evidence covers this exact commit.

### Advisory findings (this repo)

1. **Cross-repo log-string coupling**: the cli-core driver classifies this provider's warning messages by prefix. A future reword of any of the four templates in `PluginCredentialProvider.cs` (:347/:370/:374/:425) silently kills the C# loss-attribution channel; the failure surfaces later as a false-RED green run in the *other* repo's suite. Consider pinning the templates with a small unit test, or centralizing them as constants.
2. **Plant composition asymmetry (minor)**: `hammer-nolock` never constructs a provider, so C# nolock workers record no losses by construction. Plant 2's discrimination still stands (the TS nolock workers keep a live recording channel and their raced hits genuinely record nothing), but it is worth a comment in `Program.cs` so a future reader does not mistake the silence for evidence.
3. **Twin-branch lifecycle**: the suite's CI home tries `worktree-x2-concurrency-suite` before `main` forever — make sure this branch is deleted at merge, or the suite silently pins this commit as the twin while appearing green.

IvanMurzak:
## G-SEC-2 release security review (b4/c4) — verdict: PASS

Security review of the **combined b4 release surface**: `aefc8998..26fc5fe` (merges of #202 #203 #204 #205 #206 #207 #208), read at the tip. Reviewer produced none of this code. Verdict gates the NuGet publish.

**Verdict: PASS — no blocking findings.** Two cross-repo advisories (also posted on AI-Game-Dev-CLI-Core #11), one confirmation note.

### What was checked (per lens)

1. **Credential-at-rest.** `MachineCredentialStore.cs`: all three DPAPI instance sites (`TryRead` :184, `WriteProtected` :347, `ReadPlaintextJson` :315) are gated symmetrically on `IsWindows && !_plaintextForTesting`; the public ctor pins `plaintextForTesting: false`, the escape is `internal` (InternalsVisibleTo = Tests + RefreshHarness only, and IVT is not treated as a security boundary). Static `Protect/UnprotectBytes` seams always use real DPAPI. POSIX 0600/0700 applied before any secret byte lands (`WriteFileAtomic` :372); temp siblings are `CreateNew`/`O_EXCL` with unpredictable names, same-directory by contract. No plaintext write path was added anywhere in the delta.
2. **Token leakage.** Every `Unreadable` reason string is token-free (fs/DPAPI/System.Text.Json messages carry paths, Win32 codes, and positions — never document content). No log call or string concatenation in `McpPlugin/src` interpolates token material (grep-verified). `HttpTokenRefresher`/`TokenExchangeClient` failure reasons carry only server `error`/`error_description`. The x2 RefreshHarness `WARN`/`TOKEN` stdout lines print provider log messages (reason-only) and **fake-AS** tokens respectively — test infrastructure, never packed.
3. **Lock protocol** (`MachineCredentialLock.cs`). Exclusive-create only; per-attempt 128-bit nonce defeats content-identity confusion; stale takeover serialized through the intent file with candidate re-validation (mtime+size+bytes) AND intent-ownership re-verify immediately before the unlink; foreign-hostId 24 h bar; unparseable = short bar (B2, owner-adopted); budget exhaustion → busy, never lock-free; guarded release. Symlink pre-plant on lock/intent/temp paths fails closed (`CreateNew` refuses to follow); deletes remove the link, not the target. A hostile same-user process can wedge/steal the lock — availability-only by design (07 threat table, owner-settled).
4. **Refresher (b3).** F7 subject guard fires under BOTH holds in `MachineCredentialLoginCommit` with the exact-premise check (`GuardPremiseChanged`) inside the hold; hold-2 re-verifies subject, refuses a store that went missing (`StoreSignedOut` — never recreated) and treats unreadable as retryable (`StoreUnreadable`, nothing written, nothing revoked). Provider: unreadable mid-refresh = busy (never overwrite/delete); missing = adopt signed-out; `invalid_grant` requires the post-failure re-read before the dead-family verdict; transient failures never revoke or delete; orphan revocation only on terminal hold-2 aborts, outside the critical section. The ratified C#-keeps/TS-revokes unreadable asymmetry: the C# direction keeps the mint in process memory only — no on-disk or replayable artifact for a local attacker.
5. **Supply chain / package contents.** Zero new dependencies (`McpPlugin.csproj` delta = one `InternalsVisibleTo` line; DPAPI is P/Invoke by design). `deploy.yml` packs only Common/McpPlugin/Server per-project; both harness projects are `IsPackable=false` — RefreshHarness and LockHarness cannot ship. Golden-vector JSONs are `None` items (not packed) and contain only synthetic values (`plugin-access-token-A2`-style, DCR-shaped ids; no `eyJ…` material), byte-identical to the cli-core copies. `release.yml` delta is a trx-logger rename only.
6. **Public API.** `ITokenRefresher`'s DIM (owner-settled) delegates to the legacy two-string API for non-migrated engine refreshers — status-quo component-default `clientId`, transition-scoped until d1/e1/f1; it never *disables* any verification. `PluginCredentialProvider` now auto-creates the cross-process lock when none is injected — safety ON by default on upgrade. New `HttpTokenRefresher` omits `scope`/`resource` entirely (fixes the P0-3 narrowing class for adopters).

### Advisory findings (non-blocking)

- **ADV-1 (parity, both repos):** an **empty/blank `credentials.json`** classifies differently across the twins: C# `MachineCredentialStore.cs:180-181/:203-204` → `Unreadable` ("sign in required (store unreadable)", `Rotate` refuses), TS `machine-credentials.ts:352-354/:370-372` → `missing` (signed-out; the legacy TS `rotate()` would recreate a store C# refuses to touch). No token exposure, no replay, and the never-recreate-after-signout property is unaffected (sign-out deletes the file; a zero-byte file is not a sign-out artifact) — but it is a divergence in the 04 §1 unreadable-state contract on a state a killed foreign writer can produce, and x1's vectors do not cover it. Suggest aligning (either direction) or pinning the chosen asymmetry in a vector.
- **ADV-2 (CI, lives in cli-core):** the x2 `concurrency-suite.yml` clones this repo at a **mutable branch ref** (`worktree-x2-concurrency-suite` → fallback `main`) and executes the build in cli-core's CI. Blast radius is bounded (`permissions: contents: read`, no secrets), but a SHA pin (or main-only now that the pair has merged) would close the mutable-ref execution path.

**Confirmation note (not a new finding):** `PluginCredentialProvider.Adopt()` remains a machine-lock-free, subject-guard-free store write at the tip — byte-identical semantics to the baseline (`aefc8998`), i.e. pre-existing status quo whose remediation is the designed engine migration to `MachineCredentialLoginCommit` (d1/e1/f1, D6 staging per 07). Not introduced by this release; recorded so the cascades don't lose it.

PR #210, opened 2026-08-17
feat(server): server-side MCP replace-by-identity rule (W5.4 / D10.2)

IvanMurzak:
## Summary

Taskflow task **W5.4** (`.taskflow/2026-08-11-gamedev-app-ux-v2/tasks/w5-4-mcp-replace-by-identity.md`, `software` repo). Gate **`G-SERVER-MCP`** — a different gate, repository and deploy from `G-SERVER-CACHE`; approving one says nothing about the other.

Gives the MCP server a **replace-by-identity** rule: when an installation that already has a live session connects again with the same identity, the previous session is **replaced** rather than **stacked**.

This is the half of D10 that reaps sessions orphaned by a **crashed or force-killed** app. Those never send a `DELETE` — by construction, so no client-side change can ever reach them — and today they survive until the production idle timeout of **21600 s = 6 h** (`cloud/AI-Game-Dev-Server/docker-compose.yml:132`), which is the six-hour pile-up the owner reported.

---

## 1. The termination mechanism — established, and it is NOT what the obvious design assumes

**DoD 1.** The task flagged this as a hypothesis to settle before designing the rule. Settled by reading the SDK and by test, and the answer inverts the intuitive design.

Verified against `ModelContextProtocol.AspNetCore` **1.4.0** (`1.4.0+06e3604dd12fbfc08c8f9d316325380ec2a2989b`; read at that commit SHA and cross-checked by reflecting over the shipped IL in `lib/net9.0/`):

| Fact | Consequence |
|---|---|
| Session storage is **not** `StreamableHttpHandler._sessions` any more. It is `internal sealed class StatefulSessionManager` holding `ConcurrentDictionary<string, StreamableHttpSession> _sessions`. `HttpMcpSession<T>`, `TryRemoveSession` and `ActiveSessions` **do not exist** in 1.4.0. | The shape named in the task description is from an older SDK. |
| **Nothing in the SDK attaches a continuation to `ServerRunTask` that removes a session when it completes.** | ⚠️ **Returning early from `RunSessionHandler`, or cancelling the token it was handed, does NOT evict the session.** The entry survives in `_sessions` until `IdleTimeout` or `MaxIdleSessionCount` prunes it. |
| The only operation that evicts is the one the SDK's own `DELETE` handler performs: `StatefulSessionManager.TryRemove(id, out session)` then `session.DisposeAsync()`. | That is what this PR does. |
| `TryRemove`/`TryGetValue` are `public` **on an `internal` type**; `StatefulSessionManager` is a DI singleton (`TryAddSingleton` in `WithHttpTransport`); there is no `InternalsVisibleTo`. | The manager is resolved from DI **by reflection on the type handle** (`IServiceProvider.GetService(Type)` does not require a public type). |
| `internal sealed class StreamableHttpSession : IAsyncDisposable` — the interface is **public**. | Only the *lookup* needs reflection. The dispose goes through public `IAsyncDisposable`. |
| `StreamableHttpSession.DisposeAsync` disposes the transport, cancels `_disposeCts` (== the `SessionClosed` token handed to `RunSessionHandler`), awaits `ServerRunTask`, then disposes the `McpServer`. `StreamableHttpServerTransport.DisposeAsync` completes the incoming channel, releases the pending SSE `GET`, and disposes the `SseEventWriter`. | Cancelling our handler is a **consequence** of the dispose, not a substitute for it — and it is also what removes the `IMcpSessionTracker` row, since the row is removed in that handler's `finally`. |

**The trap this avoided is exactly the one the task warned about.** A rule built on "cancel our own token and let the tracker row go" would have dropped the row, looked fixed, and left the SDK session and the `SseEventWriter` buffer in place — the leak untouched, with no functional symptom. **A test asserting only "the tracker row disappeared" would have been GREEN for that broken implementation.**

### How the falling session count is proven

The load-bearing assertion is one no in-process fake can make: after a replace, a POST bearing the **replaced** session id is answered **`404` with JSON-RPC `-32001`** by the SDK's own session lookup. That is only possible if the SDK's store no longer holds the id.

`McpReplaceByIdentityOverHttpTests.SecondConnection_SameAccountAndInstance_ReplacesTheFirst_AndTheSdkReleasesIt`, over a real loopback Kestrel host with the real OAuth resource-server path, asserts in order:

1. the first session is **usable** before the replace (`ping` → 200) — otherwise "the old id 404s" could be satisfied by an id that never worked;
2. the second `initialize` mints a **different** session id;
3. the old id → **404**, and the body contains **`-32001`** (so it is "Session not found", not a routing 404);
4. the new id → **200**;
5. `IMcpSessionTracker.ActiveSessionCount` is back to **1**, not 2.

### The reflection dependency is loud, and it is a tripwire

Reflection on an internal type is a real risk, so it is made **detectable rather than latent**:

- The binding resolves **once** per process, by name **and parameter shape** (`bool (string, out …)`), anchored on the public `HttpServerTransportOptions` assembly so the assembly reference is a compile-time fact.
- Failure surfaces as `IsSupported == false` + a `BindingDiagnostic` string, and is logged **once at `Error`** — the same reasoning as `SG-15`'s loud `405`: a silent degradation here restores the six-hour pile-up with no symptom.
- **`SdkEvictionSeam_IsBoundToTheRunningSdk` asserts the binding.** An SDK bump that moves this seam **fails CI** instead of failing production. Plant **P9** proves that test discriminates (it reddens the tripwire *and* the functional replace test).
- A failed binding degrades to today's behaviour (predecessor lingers to the idle timeout); it never fails the new session's `initialize`.

---

## 2. The key — and a deliberate, documented deviation from the task spec

**DoD 2.** The slot is keyed on **`(accountSub, instanceId, projectPin)`**.

`accountSub` comes from **`ConnectionIdentity.FromPrincipal(context.User)`** — the validated credential — never from a client-supplied body field. So a client-supplied (and therefore spoofable) `instanceId` can only ever select among sessions *inside* the account the bearer token already proved: **a spoofed value evicts only the spoofer's own sessions.**

> **⚠️ Deviation, stated plainly: the task spec says `(accountSub, instanceId)`; `07-security.md` says `(accountSub, instanceId, pin)`.**
>
> - Task spec DoD 2: *"Replace is keyed on `(accountSub, instanceId)`"*.
> - `07-security.md` §2.5 copied-installation row, threat **T25**, gate **SG-14**: *"The server rule replaces on `(accountSub, instanceId, **pin**)` — two clones working on different projects then never collide."*
>
> **I implemented the three-part key**, because it is a **strict superset** that satisfies the task's DoD as written *and* honours the design:
> - same account + same installation + same project → **still replaces** (DoD 4 ✔, test + plant P1);
> - two different accounts, same instance id → **still isolated** (DoD 2 ✔, test + plant P3);
> - two clones of one copied home directory on **different projects** → no longer evict each other forever (**SG-14** ✔, test + plant P4).
>
> Dropping the pin would satisfy the task and **violate the design**, producing the permanent mutual-eviction flap T25 describes. Flagging it rather than silently choosing.

---

## 3. Released clients are unaffected — no synchronised app release

**DoD 3, DoD 6.** **The container can be deployed with every released client still on the old behaviour, and nothing regresses.** Every currently released app sends no `AI-Game-Dev-Instance-Id` header, so:

- `InstanceIdentityHeader.TryRead` returns **`Absent`** — a supported state, distinct from `Malformed`;
- `IMcpSessionReaper.Claim` is never reached, so no slot is claimed, nothing is replaced, nothing is rejected, and **nothing is logged**;
- sessions stack exactly as they do today.

`NoInstanceIdHeader_SessionsStackExactlyAsToday_AndNothingWarns` drives three header-less sessions on one account, proves all three still serve, pins `ActiveSessionCount == 3` and `TrackedIdentityCount == 0`, and asserts **no warning or error from the reaper / evictor / transport categories**.

Two things about that test worth reading:

- The warning assertion is **category-scoped**. An earlier unscoped "no warnings at all" version failed on pre-existing `tools/list` retry warnings that have nothing to do with this change — it was asserting a property of the whole test environment.
- It carries an **anti-vacuity guard**: the sink is proven non-empty, and a probe warning is emitted through the host's real `ILoggerFactory` and asserted visible. A dead sink would otherwise satisfy "empty" while the code logged a storm.

The header can also **never become mandatory**: `Absent` is a first-class state in the tri-state enum, and the compatibility path is asserted by tests and by plants P2 and P10.

---

## 4. Opaque-token validation

**DoD 7, DoD 9.** The header name is used **exactly**: `AI-Game-Dev-Instance-Id`, **no `X-` prefix** (RFC 6648), pinned as a test (`HeaderName_IsExactlyTheContractedName_WithNoXPrefix`) against both the literal and `Consts.MCP.Server.Headers.InstanceId`, because the name is half of a cross-repository contract with `w3-3` and a mismatch **fails open** (header unseen → no replace → the pile-up, silently).

### The HMAC contract, recorded verbatim for `w3-3`

```
mcpInstanceId = HMAC-SHA256(key = installId, msg = accountSub)   truncated to 128 bits, hex
```

- **key = the SECRET `installId`** (128-bit CSPRNG, persisted in `~/.ai-game-dev/config.json`)
- **msg = the PUBLIC `accountSub`**
- truncated to **128 bits** ⇒ 16 bytes ⇒ **exactly 32 hex characters** on the wire
- **Do NOT invert the key/message order.** It was independently re-verified twice and is deliberate; inverting it would require a 128-bit brute force to undo.

The server validates the value as an **opaque token** and nothing more: exactly 32 characters, hex charset, sent **at most once**. Rejected otherwise.

| Input class | Result |
|---|---|
| Absent | `Absent` — today's behaviour, untouched |
| 32 hex (any case) | `Valid`, lower-cased (an un-normalised key would give one installation two slots) |
| Over-length (33, 64) / under-length (31, 1) | **`Malformed`** |
| Non-hex at start / middle / end | **`Malformed`** |
| Contains CR, LF, CRLF, NUL, TAB | **`Malformed`** |
| Leading / trailing whitespace | **`Malformed`** — rejected, **not trimmed** |
| Present but empty | **`Malformed`** |
| Header sent **twice** | **`Malformed`** — not "first wins" |
| `../../etc/passwd…`, `' OR 1=1 --…` at valid length | **`Malformed`** |

**A present-but-malformed value is REJECTED with `400 invalid_instance_id`, not ignored** — `McpSessionTokenMiddleware`, immediately after the existing malformed-pin gate and for the same fail-closed reason. Silently ignoring it would leave a client that believes it participates in the replace rule stacking sessions forever, with no symptom short of the 6 h pile-up. The rejection is **terminal** (no handler, hub, tracker or log formatter sees the value) and **does not echo the offending value** into the response body.

**Nothing derived from the value reaches a path, a filename, or a query/LINQ predicate.** It is used for exactly two things: as a dictionary key component in `McpSessionReaper`, and as input to `Fingerprint`. Because it is validated to a closed charset *before* first use, every downstream use is safe **by construction** rather than by remembering to escape.

**And nothing reaches a log line — including on the REJECTION path.** That distinction matters: the rejection path is the one carrying *arbitrary* client input, so it is the one where an unescaped interpolation would actually be a log-injection primitive. `MalformedInstanceId_IsRejectedWith400_AndNoSessionIsEstablished` captures every record at `Trace` and asserts the malformed value appears in no log record, with a sink-liveness guard. Plant **P11** confirms it discriminates — logging the raw value there reddens all five cases. Asserting log hygiene only for values that already *passed* validation would have covered the safe half and left the dangerous half unasserted.

CR/LF cases are asserted in `InstanceIdentityHeaderTests` where the value can be handed in directly. **Every CR/LF case is padded to exactly 32 characters, with the length asserted inline**, so the **charset** check — not the length check — is what rejects it. Unpadded inputs would have been rejected on length and would have proven nothing about CR/LF handling; plant **P6** confirms this (removing the charset loop reddens 18 tests). Over the wire Kestrel refuses illegal header bytes at the protocol level, so `InstanceIdContainingCrLf_NeverProducesAWorkingSession` asserts the *outcome* rather than a specific status code.

---

## 5. Where the id is logged, and in what form

**DoD 8.** **The raw instance id is never written to any log line, and never appears beside the bearer credential.**

What *is* logged is `InstanceIdentityHeader.Fingerprint(instanceId)`: the **first 8 hex chars of `SHA-256(instanceId)`** — 32 bits, deliberately collision-prone. Enough to correlate two lines from one process and tell two installations apart during a reconnect flap; not enough to be a durable identifier in a log aggregator. It is asserted to differ from the id, not to be a substring of it, to be stable per id, and to differ across ids.

| Site | Level | Contents |
|---|---|---|
| `McpSessionReaper.Claim` — slot claimed / re-claimed | `Debug` | fingerprint, pin, MCP session id, slot count |
| `McpSessionReaper.Claim` — **session replaced** | `Information` | fingerprint, pin, replaced + new MCP session id, eviction outcome, slot count |
| `McpSessionReaper.Release` | `Debug` | fingerprint, MCP session id, released/displaced flags |
| `McpSdkSessionEvictor` — binding unsupported | `Error` (once) | binding diagnostic only (type/member names) |
| `StreamableHttpTransportLayer` — displaced cleanly | `Debug` | MCP session id only |

No credential appears on any of these lines. The value is **not** added to webhook payloads (`OnAiAgentConnected` metadata is untouched) and not to `McpClientData`.

`Replace_NeverWritesTheInstanceIdOrTheBearerTokenToALogLine` captures **every** record at `Trace` through a real MEL provider and asserts the raw id and the bearer token are absent while the fingerprint is present — and first asserts the replace actually ran, so it cannot pass by logging nothing. Plant **P7** confirms it discriminates.

---

## 6. RETENTION LIFECYCLE OF THE STORED HMAC — the answer for `w5-6`

**DoD 10.** **The honest answer is that there is no retention, and therefore no retention period to state.**

| Question | Answer |
|---|---|
| Where is it stored server-side? | **Process memory only** — one `ConcurrentDictionary` inside the `McpSessionReaper` singleton (`McpPlugin.Server/src/McpSessionReaper.cs`). |
| Is it persisted? | **No.** No database, no disk, no cache, no Redis, no webhook payload, no `McpClientData`, no log line (logs carry only the 32-bit non-reversible fingerprint). |
| For how long does it survive? | Exactly the lifetime of the MCP session that presented it. |
| What removes it? | (a) the session ending — the lease is disposed in the same `finally` that removes the session's `IMcpSessionTracker` row, covering client `DELETE`, idle-timeout eviction, transport loss and host shutdown; (b) a newer session with the same identity replacing it; (c) process exit / container restart. |
| Does it outlive the session it keys? | **No.** Asserted by `DisposingTheLease_ReleasesTheSlot_SoNothingOutlivesItsSession`. |
| Any bounded retention period? | **None, because nothing is retained.** There is no store to expire and no sweep to schedule. |
| Residual caveat, stated rather than omitted | If the container runs at `Debug`/`Information` level, the **fingerprint** (not the value) appears in the container log, subject to that log's normal `json-file` rotation. The identifier itself never does. |

**`w5-6` (PR IvanMurzak/AI-Game-Dev-Server#599) is already MERGED, and it correctly did not invent a period** — its body states *"No retention period is stated. The server-side HMAC retention lifecycle is unverified pending `w5-4-mcp-replace-by-identity`."* That was the right call and the answer above confirms it needs no amendment: a policy that states purpose and storage location, and claims no period, is accurate for this implementation. **This answer has been posted as a comment on #599** so the record lives where `w5-6` is, not only here.

---

## 7. Delivery chain — hop by hop, and a correction

**DoD 11.** ⚠️ **This change is NOT in production when this PR merges.** It is in production only when hop 4 lands.

| Hop | What | Owner | Status |
|---|---|---|---|
| **1** | `MCP-Plugin-dotnet` merge to `main` → `release.yml` (version read from `McpPlugin/McpPlugin.csproj`, now **8.2.0**) → GitHub release → `deploy.yml` on `release: released` → `dotnet nuget push` of **three** packages (`McpPlugin.Common`, `McpPlugin`, `McpPlugin.Server`) | **this PR, automatically on merge** | ⏳ **performed by merging this PR** |
| **2** | Bump the consuming pins from 8.1.0 → **8.2.0** — see the correction below | orchestrator / owner, **after 8.2.0 is on nuget.org** | ❌ remains |
| **3** | `GameDev-MCP-Server` release → `deploy_docker.yml` on `release: published` → push `aigamedeveloper/mcp-server:<new tag>` to Docker Hub | orchestrator / owner | ❌ remains |
| **4** | `cloud/AI-Game-Dev-Server/docker-compose.yml:64` — bump `aigamedeveloper/mcp-server:9.2.5` to the new tag. **Merging that to `main` IS the production deploy** (full restart, ~4 min, expected nginx 502 window). | orchestrator / owner | ❌ remains |

> ### ⚠️ Correction to the task spec's hop 2 — there are TWO pin sites, not one
>
> The task spec names only `shared/GameDev-MCP-Server/com.IvanMurzak.GameDev.MCP.Server.csproj:60,67`. Verified against the repositories, there is a **second** consumer, and the last real bump (`AI-Game-Dev-Server#597`, for 8.1.0) touched only that one:
>
> | Pin site | Role | Currently |
> |---|---|---|
> | `shared/GameDev-MCP-Server/com.IvanMurzak.GameDev.MCP.Server.csproj:60,67` | **The production path** — builds `aigamedeveloper/mcp-server` | 8.1.0 |
> | `cloud/AI-Game-Dev-Server/mcp-server/McpServer.csproj:26,27` (+ the `ModelContextProtocol.AspNetCore` "tracks the transitive version" comment) | Vendored **reference/fallback** source, built only by `docker-compose.local.yml`. Not production — prod pulls the published image. | 8.1.0 |
>
> Production is gated by the **first** row only, so the second cannot block the deploy. But `#597` bumped it in lockstep, and leaving it behind makes the local/fallback build silently diverge from prod. **Bump both.**
>
> Also stale in the task spec: it states all three packages and the `GameDev-MCP-Server` pins are at **8.0.0**. They were at **8.1.0** (HEAD is `1e2eda1 chore: release MCP-Plugin-dotnet 8.1.0 (#209)`); this PR moves them to 8.2.0.

**DoD 12 — version bumps are deliberate and lockstepped.** All three moved together via `commands/bump-version.ps1`, the repo's single lever (it also updates `server.json` and `Consts.PluginVersion`). 8.1.0 → **8.2.0**, minor because this adds a header contract and new public seams without changing behaviour for any client that does not send the header.

> **A re-push of an already-published NuGet version is a SILENT no-op** under `dotnet nuget push --skip-duplicate`. Tag `8.1.0` already exists, so **without this bump, merging would have produced no release, no publish, and no way to ever ship this rule under 8.1.0.** A NuGet publish is a third irreversible publication class: 8.2.0 cannot be replaced once pushed.

---

## 8. Rollback — and the SG-5 coupling

**DoD 13.** Reverse order: revert the rule → publish a new version (**never re-push 8.2.0**; it is irreversible and `--skip-duplicate` would no-op) → bump both pin sites back → rebuild and re-tag the image → bump `docker-compose.yml:64` back and merge (another full restart).

> ### ⚠️ **`w3-3` must be held or reverted WITH this — do not revert one half alone.**
>
> Under **SG-5**, *an identifier ships only together with a server-side consumer, or not at all*. `w5-4` **is** that consumer. Reverting the server rule while `w3-3` still ships the identifier leaves a persisted device identifier transmitted to a server that does nothing with it — no benefit, and it re-opens the `G-PRIVACY` exposure that `w5-6` was merged to close. Conversely, `w3-3` merging before this reaches production merely means the header is sent and ignored, which is harmless and is the intended ordering.
>
> Cheaper mitigation before reaching for a rollback: the rule is **inert for any client that does not send the header**, so holding or reverting `w3-3` alone disables it in practice without touching this repository.

**DoD 14 — `G-SERVER-MCP`.** Respected. This is a different gate, a different repository and a different deploy from `G-SERVER-CACHE` (which covers `w5-1`/`w5-3`/`w5-5`/`w5-6` in `cloud/AI-Game-Dev-Server`). Approving one says nothing about the other.

---

## 9. Test plan and the plant round

**DoD 15.** Full solution build and test suite, on the final tree, both target frameworks:

```
dotnet build McpPlugin.sln          -> Build succeeded, 0 errors
                                       12 warnings == the 12 on the baseline commit; none in changed files
dotnet test  McpPlugin.sln
  McpPlugin.Server.Tests  net8.0 -> Passed: 745, Failed: 0
  McpPlugin.Server.Tests  net9.0 -> Passed: 745, Failed: 0
  McpPlugin.Tests         net8.0 -> Passed: 943, Failed: 0
  McpPlugin.Tests         net9.0 -> Passed: 943, Failed: 0
```

The new integration tests bind loopback ports and run real Kestrel hosts, so the CI OS matrix is load-bearing rather than incidental: **`test (ubuntu-latest)` and `test (macos-latest)` both pass**, which is the evidence that the eviction seam and the real-host harness are not Windows-only.

Baseline on `1e2eda1` before any change was `McpPlugin.Server.Tests` **684** passing on both TFMs, so **+61 new tests** and no pre-existing failures to disentangle.

### Plant round — 13 plants, 13 RED

**DoD 5** asks for two plants. Thirteen were run, because the unit of verification is a **claim**, not a test, and `N/N RED` for a suite says nothing about which claim each test can actually fail on. Each plant was applied as an exact source edit against the verified-green commit, tested, and reverted (`git status --porcelain` clean after every one).

| # | Break planted | Claim it attacks | Verdict | Tests reddened |
|---|---|---|---|---|
| **P1** | Reaper never calls `Evict` | **DoD 4 + 1** — a second connection replaces the first | 🔴 RED | 7 |
| **P2** | `Claim`'s null-identity guard removed (empty-string pooling) | **DoD 3** — the rule must not fire on a null identity | 🔴 RED | 2 |
| **P3** | `accountSub` dropped from the key | **DoD 2** — cross-account isolation | 🔴 RED | 3 |
| **P4** | `projectPin` dropped from the key | **07 T25 / SG-14** — copied-installation flap | 🔴 RED | 2 |
| **P5** | Middleware ignores a malformed header instead of rejecting | **DoD 7** — malformed is rejected, not ignored | 🔴 RED | 5 |
| **P6** | Hex charset loop removed (length only) | **DoD 7** — the *charset* gate rejects CR/LF and payloads | 🔴 RED | 18 |
| **P7** | Log the raw instance id instead of the fingerprint | **DoD 8** — the id never reaches a log line | 🔴 RED | 1 |
| **P8** | `Release` made an unconditional `TryRemove(key)` | a displaced session must not delete its successor's slot | 🔴 RED | 2 |
| **P9** | SDK type name changed (binding broken) | **DoD 1** — the SDK-bump tripwire fires | 🔴 RED | 3 |
| **P10** | Transport treats **every** session as carrying an identity | **DoD 3 at the HTTP level** | 🔴 RED | 3 |
| **P11** | Log the raw client value on the **rejection** path | **DoD 7/8** — the id never reaches a log line, incl. rejected input | 🔴 RED | 5 |
| **P12** | `instanceId` dropped from the key | one account's two installations must stack (desktop + laptop) | 🔴 RED | 2 |
| **P13** | Flag the outgoing lease on a same-session re-claim | a re-claim must mis-attribute nothing | 🔴 RED | 1 (exactly the target) |

**Restored → GREEN**: 745/745 both TFMs (§9 above).

> **P10 exists because P2 was not sufficient, and that is the interesting result.** P2 broke `Claim`'s null-identity guard and reddened only the **unit** tests — the HTTP compatibility test `NoInstanceIdHeader_SessionsStackExactlyAsToday_AndNothingWarns` **stayed green**, because over HTTP the transport never calls `Claim` at all when the header is absent, so P2 could not reach that path. A test that survives the break aimed at it has not been shown to discriminate. **P10 breaks it where it actually lives** — making the transport treat every session as carrying a valid identity, so header-less clients pool onto one slot per account and evict each other — and the DoD-3 HTTP test goes red. Without P10, DoD 5's second half would have been proven only at the unit level while reading as fully proven.

### Three defects found before review

- `ConcurrentClaimsForOneIdentity_ConvergeOnASingleSlot` flaked in a combined run and passed in isolation. Cause was the **test's own** `List<string>` recorder being unsynchronised — the reaper calls `Evict` outside its claim lock on purpose (never hold a lock across foreign code), so concurrent claims reach the fake concurrently. Read as a race in the production code and was not one; the recorder is now locked.
- An unused `InstanceY` constant in the HTTP suite looked like dead code and was actually a **coverage gap**: nothing exercised one account with two installations — the everyday desktop-plus-laptop case, and the most user-visible way this rule could go wrong. Adding the test (rather than deleting the constant) closed it; plant **P12** confirms it discriminates.
- **A real defect in the shipped code, found by re-reading the diff as a reviewer would.** The same-session-re-claim check ran *after* `MarkDisplaced`, so a re-entrant claim flagged the **outgoing** lease `Displaced` and only then discovered it was the same session. No session was evicted — that part was correct — but the flag made that session's handler report an ordinary cancellation as *"displaced by replace-by-identity"*. **Wrong attribution is worse than none:** it sends an operator chasing a replace that never happened. The existing test asserted only the *incoming* lease's flag, which is exactly why it missed it. Both the check and `MarkDisplaced` now live inside the claim lock (which also deleted a dead `!ReferenceEquals(existing, lease)` guard that could never be false), the test asserts both leases, and it was confirmed by running the new assertion **against the unfixed code**, where it fails with its own message — stronger evidence than a plant, because the defect was real rather than injected.
- The DoD-3 warning assertion was originally unscoped (`no warnings at all`) and failed on pre-existing `tools/list` retry noise. It also drove liveness with `tools/list`, which with no engine plugin attached retries ten times per call — 10 s per probe, 10 warnings and an error each. Switched to `ping` (handled inside the MCP session, so it cleanly discriminates live-vs-evicted): the HTTP suite went from ~100 s to ~1.5 s and the log assertions became meaningful.

---

## 10. Files changed

**New**

| File | Purpose |
|---|---|
| `McpPlugin.Server/src/Auth/InstanceIdentityHeader.cs` | Header name, tri-state opaque-token validation, log fingerprint |
| `McpPlugin.Server/src/IMcpSessionReaper.cs` | The rule's contract, key shape, and the retention statement |
| `McpPlugin.Server/src/McpSessionReaper.cs` | In-memory slot registry; claim-and-displace; ABA-safe release |
| `McpPlugin.Server/src/IMcpSdkSessionEvictor.cs` | Eviction seam; splits the synchronous leak-stopping half from the async dispose |
| `McpPlugin.Server/src/McpSdkSessionEvictor.cs` | `TryRemove` + `DisposeAsync` against the SDK, with the loud binding diagnostic |
| `McpPlugin.Server.Tests/InstanceIdentityHeaderTests.cs` | 38 tests — validation, header reads, fingerprint |
| `McpPlugin.Server.Tests/McpSessionReaperTests.cs` | 11 tests — the key algebra against a recording fake evictor |
| `McpPlugin.Server.Tests/McpReplaceByIdentityOverHttpTests.cs` | 12 tests — real Kestrel + real OAuth + real MCP handshake |

**Modified**

| File | Change |
|---|---|
| `McpPlugin.Common/src/Utils/Consts.MCP.cs` | `Headers.InstanceId = "AI-Game-Dev-Instance-Id"` |
| `McpPlugin.Server/src/Auth/McpSessionTokenMiddleware.cs` | Fail-closed instance-id gate (`400 invalid_instance_id`) beside the pin gate |
| `McpPlugin.Server/src/Transport/StreamableHttpTransportLayer.cs` | Claim the slot in `RunSessionHandler`; bounded wait on the predecessor's dispose; release the lease in `finally`; orderly cancellation logged at `Debug` not `Error` |
| `McpPlugin.Server/src/Extension/ExtensionsMcpServerBuilder.cs` | Register both singletons in all modes |
| 5 files | Lockstepped 8.1.0 → 8.2.0 |

### Three implementation details worth a reviewer's attention

1. **`Release` is a compare-and-remove, not `TryRemove(key)`.** A displaced session unwinds *after* its successor installed itself, and its `finally` disposes its lease. An unconditional removal would delete the **successor's** slot, stranding a live session that no later connection could find and therefore never replace — the leak back, reachable only through the code added to fix it, with every other test still passing. Plant **P8** covers it.
2. **The wait on the predecessor's dispose is bounded (10 s), and that is safe.** `TryRemove` — the leak-stopping half — completes **synchronously inside `Claim`**, so a timeout cannot leave a reachable session or reintroduce the pile-up. The wait exists only so the common case is fully settled (and deterministic for tests) before the replacement starts serving. Unbounded would let one session's teardown, which awaits a SignalR disconnect notification, hang another's `initialize`.
3. **The rule can never prevent a session from starting.** The whole claim block is wrapped: any failure logs and degrades to today's behaviour. A client that cannot connect is strictly worse than one whose predecessor lingers until the idle timeout.

---

## Out of scope, untouched

`w3-3` (app-side identifier), `w3-5` (Settings reset), `w3-1` (client `DELETE`), the SG-3 verification read, and **any change to `IdleTimeout` or `MaxIdleSessionCount`** — both still read from `DataArguments` exactly as before.

https://claude.ai/code/session_014wEWRDnWNQMPgKEiTz7SMj

### DIFF
--- a/McpPlugin.Common/McpPlugin.Common.csproj
+++ b/McpPlugin.Common/McpPlugin.Common.csproj
@@ -11,7 +11,7 @@
     <!-- NuGet Package Information -->
     <IsPackable>true</IsPackable>
     <PackageId>com.IvanMurzak.McpPlugin.Common</PackageId>
-    <Version>8.1.0</Version>
+    <Version>8.2.0</Version>
     <Authors>Ivan Murzak</Authors>
     <Company>Ivan Murzak</Company>
     <Copyright>Copyright © Ivan Murzak 2025</Copyright>
--- a/McpPlugin.Common/src/Utils/Consts.MCP.cs
+++ b/McpPlugin.Common/src/Utils/Consts.MCP.cs
@@ -274,6 +274,23 @@ namespace com.IvanMurzak.McpPlugin.Common
 
                     /// <summary>Value the trusted-client header must carry to opt in.</summary>
                     public const string TrustedInternalClientOptInValue = "1";
+
+                    /// <summary>
+                    /// Per-installation MCP instance identity (design 07 §2.3, D10.2). The value is
+                    /// <c>HMAC-SHA256(key = installId, msg = accountSub)</c> truncated to 128 bits and
+                    /// hex-encoded — i.e. exactly 32 lower-case hex characters. It is stable per
+                    /// <c>(installation, account)</c> and unlinkable across accounts, because the server
+                    /// never sees <c>installId</c>.
+                    ///
+                    /// <para><b>No <c>X-</c> prefix, per RFC 6648.</b> The name must match the client's
+                    /// byte-for-byte; see <c>InstanceIdentityHeader</c> for the validation contract.</para>
+                    ///
+                    /// <para><b>This is a routing key, never an authenticator.</b> It is only ever used
+                    /// WITHIN the <c>accountSub</c> scope the bearer credential already established, so a
+                    /// spoofed value can evict only the spoofer's own sessions. Absent is a supported
+                    /// state: a client that never sends it keeps the pre-existing behaviour exactly.</para>
+                    /// </summary>
+                    public const string InstanceId = "AI-Game-Dev-Instance-Id";
                 }
             }
         }
--- a/McpPlugin.Common/src/Utils/Consts.cs
+++ b/McpPlugin.Common/src/Utils/Consts.cs
@@ -12,7 +12,7 @@ namespace com.IvanMurzak.McpPlugin.Common
     public static partial class Consts
     {
         public const string ApiVersion = "2.0.0";
-        public const string PluginVersion = "8.1.0";
+        public const string PluginVersion = "8.2.0";
 
         public static class Guid
         {
--- a/McpPlugin.Server.Tests/InstanceIdentityHeaderTests.cs
+++ b/McpPlugin.Server.Tests/InstanceIdentityHeaderTests.cs
@@ -1,0 +1,228 @@
+/*
+┌────────────────────────────────────────────────────────────────────────┐
+│  Author: Ivan Murzak (https://github.com/IvanMurzak)                   │
+│  Repository: GitHub (https://github.com/IvanMurzak/MCP-Plugin-dotnet)  │
+│  Copyright (c) 2025 Ivan Murzak                                        │
+│  Licensed under the Apache License, Version 2.0.                       │
+│  See the LICENSE file in the project root for more information.        │
+└────────────────────────────────────────────────────────────────────────┘
+*/
+
+#nullable enable
+using System;
+using System.Linq;
+using com.IvanMurzak.McpPlugin.Common;
+using com.IvanMurzak.McpPlugin.Server.Auth;
+using Microsoft.AspNetCore.Http;
+using Shouldly;
+using Xunit;
+
+namespace com.IvanMurzak.McpPlugin.Server.Tests
+{
+    /// <summary>
+    /// The opaque-token contract for <c>AI-Game-Dev-Instance-Id</c> (design 07 §2.3, D10.2).
+    ///
+    /// <para>Everything here exists because the value is <b>client-supplied and untrusted</b>. It is
+    /// validated to a fixed length over a closed charset BEFORE it is used anywhere, which is what makes
+    /// every downstream use safe by construction instead of safe by remembering to escape.</para>
+    ///
+    /// <para><b>The tri-state is the point.</b> ABSENT and MALFORMED must not collapse into each other:
+    /// absent is every currently released client and must behave exactly as it does today, while malformed
+    /// is a client that believes it is participating in the replace rule and would otherwise silently
+    /// stack sessions forever.</para>
+    /// </summary>
+    public sealed class InstanceIdentityHeaderTests
+    {
+        // A well-formed value: 128 bits of HMAC output, hex ⇒ exactly 32 chars.
+        const string ValidId = "0123456789abcdef0123456789abcdef";
+
+        [Fact]
+        public void HeaderName_IsExactlyTheContractedName_WithNoXPrefix()
+        {
+            // DoD 9 / RFC 6648. This is pinned as a test rather than trusted to review because the name is
+            // half of a cross-repository contract: the app (W3.3) must send this byte-for-byte, and a
+            // mismatch fails OPEN — the header is simply not seen, no replace happens, and the only symptom
+            // is the six-hour session pile-up the rule exists to end.
+            InstanceIdentityHeader.HeaderName.ShouldBe("AI-Game-Dev-Instance-Id");
+            InstanceIdentityHeader.HeaderName.ShouldBe(Consts.MCP.Server.Headers.InstanceId);
+            InstanceIdentityHeader.HeaderName.StartsWith("X-", StringComparison.OrdinalIgnoreCase)
+                .ShouldBeFalse("RFC 6648 deprecates the X- prefix, and the client does not send one");
+        }
+
+        [Fact]
+        public void ExpectedLength_Is32_Because128BitsHexIs32Chars()
+        {
+            // 128 bits ⇒ 16 bytes ⇒ 32 hex chars. Pinned so a "let's allow 64 too" relaxation is a
+            // deliberate, visible change rather than a quiet widening of an opaque-token contract.
+            InstanceIdentityHeader.ExpectedLength.ShouldBe(32);
+            ValidId.Length.ShouldBe(InstanceIdentityHeader.ExpectedLength);
+        }
+
+        // ───────────────────────────── accepted ─────────────────────────────
+
+        [Fact]
+        public void LowerCaseHex_IsValid_AndPreserved()
+        {
+            InstanceIdentityHeader.TryParse(ValidId, out var parsed).ShouldBe(InstanceIdentityParse.Valid);
+            parsed.ShouldBe(ValidId);
+        }
+
+        [Fact]
+        public void UpperCaseHex_IsValid_AndNormalisedToLowerCase()
+        {
+            // Normalisation matters for the KEY, not for politeness: "ABCD…" and "abcd…" are the same
+            // installation, and an un-normalised key would give one installation two slots — so its two
+            // sessions would never replace each other and the pile-up would survive for that client.
+            InstanceIdentityHeader.TryParse(ValidId.ToUpperInvariant(), out var parsed)
+                .ShouldBe(InstanceIdentityParse.Valid);
+            parsed.ShouldBe(ValidId);
+        }
+
+        // ───────────────────────────── rejected ─────────────────────────────
+
+        [Theory]
+        // Over-length — one char over, and a full 64-char (256-bit) value that is otherwise perfect hex.
+        [InlineData("0123456789abcdef0123456789abcdef0")]
+        [InlineData("0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef")]
+        // Under-length: a TRUNCATED value is the dangerous near-miss, because it still looks like hex.
+        [InlineData("0123456789abcdef0123456789abcde")]
+        [InlineData("0")]
+        // Non-hex charset, at the start, middle and end — 'g' is the first letter past 'f'.
+        [InlineData("g123456789abcdef0123456789abcdef")]
+        [InlineData("0123456789abcdefg123456789abcdef")]
+        [InlineData("0123456789abcdef0123456789abcdeg")]
+        // Structural characters that would matter if the value were ever concatenated into something.
+        [InlineData("0123456789abcdef0123456789abcd..")]
+        [InlineData("../123456789abcdef0123456789abcd")]
+        [InlineData("0123456789abcdef0123456789ab'--")]
+        // Whitespace padding is REJECTED, not trimmed: silently repairing it would make client and server
+        // disagree about what the identity is, which is how a replace rule evicts the wrong session.
+        [InlineData(" 123456789abcdef0123456789abcdef")]
+        [InlineData("0123456789abcdef0123456789abcde ")]
+        // Empty and whitespace-only.
+        [InlineData("")]
+        [InlineData("                                ")]
+        public void MalformedValues_AreRejected(string raw)
+        {
+            InstanceIdentityHeader.TryParse(raw, out var parsed).ShouldBe(InstanceIdentityParse.Malformed);
+            parsed.ShouldBeNull("a rejected value must never be handed onward in any repaired form");
+        }
+
+        [Theory]
+        // CR/LF and friends in the arrangements a log-injection or response-splitting attempt would use.
+        // EVERY case here is EXACTLY 32 characters long, so the length check cannot be what rejects it —
+        // the charset check has to do the work. That padding discipline is the whole value of this Theory:
+        // unpadded inputs would be rejected on length and would prove nothing at all about CR/LF handling,
+        // and the test would stay green with the charset loop deleted.
+        [InlineData("0123456789abcdef\r\n6789abcdef0123")]  // 16 + 2 + 14
+        [InlineData("0123456789abcdef\n0123456789abcde")]    // 16 + 1 + 15
+        [InlineData("0123456789abcdef\r0123456789abcde")]    // 16 + 1 + 15
+        [InlineData("\r\n23456789abcdef0123456789abcdef")]  // 2 + 30, leading
+        [InlineData("0123456789abcdef0123456789abcd\r\n")]  // 30 + 2, trailing
+        [InlineData("0123456789abcdef\00123456789abcde")]    // NUL
+        [InlineData("0123456789abcdef\t0123456789abcde")]    // TAB
+        public void ControlCharacters_AreRejectedByCharsetNotByLength(string raw)
+        {
+            // The discriminator, asserted inline so it cannot rot: if a future edit changed one of these
+            // inputs to the wrong length, this line fails loudly instead of the case quietly degrading into
+            // a length test that happens to pass.
+            raw.Length.ShouldBe(InstanceIdentityHeader.ExpectedLength,
+                "this case must be rejected by the CHARSET check, so it must have the VALID length");
+
+            InstanceIdentityHeader.TryParse(raw, out var parsed).ShouldBe(InstanceIdentityParse.Malformed);
+            parsed.ShouldBeNull("a rejected value must never be handed onward in any repaired form");
+        }
+
+        [Fact]
+        public void NullValue_IsMalformed_NotAbsent()
+        {
+            // TryParse is only reached when the header was PRESENT. A present header with a null value is a
+            // client that asked to participate and sent nothing — malformed, not absent.
+            InstanceIdentityHeader.TryParse(null, out var parsed).ShouldBe(InstanceIdentityParse.Malformed);
+            parsed.ShouldBeNull();
+        }
+
+        // ───────────────────────────── header dictionary reads ─────────────────────────────
+
+        [Fact]
+        public void NoHeaderAtAll_IsAbsent_WhichIsTheReleasedClientPath()
+        {
+            var headers = new HeaderDictionary();
+            InstanceIdentityHeader.TryRead(headers, out var parsed).ShouldBe(InstanceIdentityParse.Absent);
+            parsed.ShouldBeNull();
+        }
+
+        [Fact]
+        public void NullHeaderDictionary_IsAbsent()
+        {
+            InstanceIdentityHeader.TryRead(null, out var parsed).ShouldBe(InstanceIdentityParse.Absent);
+            parsed.ShouldBeNull();
+        }
+
+        [Fact]
+        public void HeaderPresentWithValidValue_IsValid_AndCaseInsensitiveOnTheNAME()
+        {
+            // HTTP header names are case-insensitive, and IHeaderDictionary honours that. Asserted so a
+            // client that sends "ai-game-dev-instance-id" is not silently treated as sending nothing.
+            var headers = new HeaderDictionary { { "ai-game-dev-instance-id", ValidId } };
+            InstanceIdentityHeader.TryRead(headers, out var parsed).ShouldBe(InstanceIdentityParse.Valid);
+            parsed.ShouldBe(ValidId);
+        }
+
+        [Fact]
+        public void HeaderSentTwice_IsMalformed_NotFirstWins()
+        {
+            // Two values mean the caller's intent is ambiguous. Taking the first would let a caller smuggle
+            // a second identity past anyone reading only the first, and would make which session gets
+            // evicted depend on header ordering.
+            var headers = new HeaderDictionary
+            {
+                { InstanceIdentityHeader.HeaderName, new Microsoft.Extensions.Primitives.StringValues(new[] { ValidId, "fedcba9876543210fedcba9876543210" }) }
+            };
+            InstanceIdentityHeader.TryRead(headers, out var parsed).ShouldBe(InstanceIdentityParse.Malformed);
+            parsed.ShouldBeNull();
+        }
+
+        [Fact]
+        public void HeaderPresentWithEmptyValue_IsMalformed()
+        {
+            var headers = new HeaderDictionary { { InstanceIdentityHeader.HeaderName, string.Empty } };
+            InstanceIdentityHeader.TryRead(headers, out var parsed).ShouldBe(InstanceIdentityParse.Malformed);
+            parsed.ShouldBeNull();
+        }
+
+        // ───────────────────────────── fingerprint ─────────────────────────────
+
+        [Fact]
+        public void Fingerprint_NeverReturnsTheValue_AndIsShortHex()
+        {
+            var fingerprint = InstanceIdentityHeader.Fingerprint(ValidId);
+
+            fingerprint.ShouldNotBe(ValidId, "the log-safe form must not be the identifier itself");
+            ValidId.Contains(fingerprint, StringComparison.Ordinal).ShouldBeFalse(
+                "the fingerprint must not be a substring of the id — a prefix would leak the id's own bytes");
+            fingerprint.Length.ShouldBe(8);
+            fingerprint.All(c => (c >= '0' && c <= '9') || (c >= 'a' && c <= 'f')).ShouldBeTrue();
+        }
+
+        [Fact]
+        public void Fingerprint_IsStableForOneIdAndDiffersAcrossIds()
+        {
+            // Both halves matter: stable makes it useful for correlating two log lines about one
+            // installation; different makes it useful for telling two installations apart. A constant would
+            // pass "stable" alone and be worthless.
+            InstanceIdentityHeader.Fingerprint(ValidId).ShouldBe(InstanceIdentityHeader.Fingerprint(ValidId));
+            InstanceIdentityHeader.Fingerprint(ValidId)
+                .ShouldNotBe(InstanceIdentityHeader.Fingerprint("fedcba9876543210fedcba9876543210"));
+        }
+
+        [Fact]
+        public void Fingerprint_OfNothing_IsAPlaceholder_NotAnException()
+        {
+            // The log path must never be able to throw: a logging failure inside a session teardown would
+            // turn a clean replace into a fault.
+            InstanceIdentityHeader.Fingerprint(null).ShouldBe("none");
+            InstanceIdentityHeader.Fingerprint(string.Empty).ShouldBe("none");
+        }
+    }
+}
--- a/McpPlugin.Server.Tests/McpReplaceByIdentityOverHttpTests.cs
+++ b/McpPlugin.Server.Tests/McpReplaceByIdentityOverHttpTests.cs
@@ -170,6 +170,46 @@ namespace com.IvanMurzak.McpPlugin.Server.Tests
                 .ShouldBe(HttpStatusCode.OK);
         }
 
+        /// <summary>
+        /// ONE account with TWO different installations — the everyday desktop-plus-laptop case — stacks.
+        /// Both sessions stay live, and each still replaces only its own predecessor.
+        ///
+        /// <para>This is the most user-visible way the rule could go wrong: a key that ignored the instance
+        /// id would make signing in on a laptop silently kill the session on the desktop, which reads as a
+        /// network fault to the user. Covered at the unit level too, but asserted here because this is the
+        /// path production actually takes.</para>
+        /// </summary>
+        [Fact]
+        public async Task OneAccountWithTwoInstallations_StacksAndEachReplacesOnlyItsOwn()
+        {
+            await using var host = await StartOAuthHostAsync();
+            var tracker = host.Services.GetRequiredService<IMcpSessionTracker>();
+            using var client = NewClient();
+
+            var desktop = await host.InitializeAsync(client, AccountA, InstanceX);
+            var laptop = await host.InitializeAsync(client, AccountA, InstanceY);
+
+            laptop.ShouldNotBe(desktop);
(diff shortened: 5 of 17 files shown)


IvanMurzak:
## Taskflow review - depth medium, gating advisory (per taskflow-execute code-review.md)

Reviewed by a separate agent from the one that produced this diff, per the reviewer-is-never-the-implementer rule. Read via gh pr diff 210 / gh pr view 210 only - no worktree checkout of this branch. At medium depth this covers DoD conformance, obvious defects, reuse/simplification, efficiency, whether the new tests genuinely discriminate, and adjacent call sites. medium gates advisory only - nothing below blocks merge by itself, but see the closing note on why that still matters for this PR specifically.

### Bottom line

No blocking or defect-level findings. This is an unusually careful diff: every claim I checked against the actual code held, the test suite is real (loopback Kestrel plus real OAuth plus real MCP handshake, not mocks), and the two self-reported defects (P2-insufficiency leading to P10; the pre-MarkDisplaced re-claim ordering bug) are both genuinely fixed and pinned by tests that fail on the unfixed code. Two non-blocking observations below, nothing more.

### Risk concentration 1 - the reflection seam / SdkEvictionSeam_IsBoundToTheRunningSdk

Verdict: the tripwire is real, not vacuous, but its coverage is narrower than "would catch any SDK regression", and that is fine because a second test covers the gap.

- McpSdkSessionEvictor's static constructor (McpSdkSessionEvictor.cs around line 1924-1964 in the diff) reflects over typeof(HttpServerTransportOptions).Assembly - the actual ModelContextProtocol.AspNetCore 1.4.0 assembly the test project references, not a stub. IsBindingResolved and StaticBindingDiagnostic are computed from that real reflection at process start, so the tripwire test asserting IsBindingResolved == true is testing a fact about the real, currently-pinned SDK, not a tautology.
- What it catches: StatefulSessionManager being renamed or removed, or TryRemove losing the bool (string, out T) shape. Plant P9 ("SDK type name changed") reddened this test plus 2 others.
- What it does NOT catch by itself: a future SDK version that keeps the exact type and method name and signature but changes behavior (e.g. TryRemove returns true without actually detaching the session's SseEventWriter, or DisposeAsync stops completing ServerRunTask). A binding-shape test cannot see that.
- That gap is covered by a different, non-vacuous test: McpReplaceByIdentityOverHttpTests.SecondConnection_SameAccountAndInstance_ReplacesTheFirst_AndTheSdkReleasesIt asserts the SDK's own 404 with -32001 on the replaced session id, and ActiveSessionCount falling back to 1 - an outcome only the SDK's real session store can produce. It also proves the old session was live BEFORE the replace, so a false-404 cannot be mistaken for a real eviction.
- Combined: binding-shape regression fails the tripwire; behavior regression with intact binding fails the functional HTTP test. "A future SDK bump silently defeats the eviction and CI stays green" is not a scenario I could construct. No finding.

### Risk concentration 2 - fail-closed identity gate / DoS via a client-supplied header

Verdict: no cross-account eviction path exists; the gate rejects self-consistently and cannot be used to touch another account's session.

- InstanceIdentityHeader.TryParse enforces exact length (32) BEFORE the charset loop, so an over-length payload never reaches per-char work. Charset is a closed [0-9a-fA-F] set, so CR, LF, NUL, TAB, whitespace, path-traversal and SQL-comment payloads all fail on the first non-hex character; verified both by direct unit tests (CR/LF cases padded to exactly 32 chars, so the charset check and not the length check is what discriminates) and by an HTTP-level theory test.
- A header sent twice is Malformed, not "first wins" - closes the obvious smuggling angle.
- Missing header means no-op. Claim returns null before touching the dictionary when accountSub or instanceId is empty, checked first specifically to avoid pooling every header-less caller onto one empty-string slot. Verified end-to-end by a test that also asserts zero warnings or errors from the reaper/evictor/transport categories, with a sink-liveness probe so "empty" cannot be satisfied by a dead sink.
- Malformed header means 400, terminal, before any identity resolution (McpSessionTokenMiddleware.cs). A malformed value can never reach Claim. Confirmed by a test asserting zero sessions and zero tracked identities after rejection.
- accountSub provenance: ConnectionIdentity.FromPrincipal(context.User) with a fallback to McpSessionTokenContext.CurrentIdentity, in StreamableHttpTransportLayer.cs - both sources are populated only from the validated bearer credential (the JWT sub / PAT-owner claim), never a request body or header field. A spoofed instanceId therefore selects only among the spoofer's own account's slots, confirmed both at the unit level and over real HTTP with two real bearer tokens for two different accounts.
- No finding. For the record: in non-OAuth/legacy auth modes the reaper is unconditionally inert too (no sub claim to resolve) - a scope boundary, not a gap.

### Risk concentration 3 - the three-part key (accountSub, instanceId, pin)

Verdict: correctly and consistently three-part everywhere the key is used; no residual two-part comparison found.

- Single definition site: McpSessionReaper.IdentityKey is a record(AccountSub, InstanceId, ProjectPin), explicitly ordinal-only (no culture or case folding), and it is the only key type in the reaper - Claim, Release (compare-and-remove on key+lease), and CurrentSessionIdFor all go through it. No second, narrower key or lookup path exists anywhere in the diff.
- The pin genuinely participates in the decision, both directions: a dedicated test proves two different pins for the same account and instance do NOT collide, and the SAME pin still replaces (the second half exists specifically so the first half cannot be satisfied by "pin present means never replace").
- The deviation from the task spec's literal (accountSub, instanceId) is disclosed, not silently substituted, and is a strict superset - every DoD-4/DoD-2 scenario I traced still holds with the pin present, and it additionally satisfies 07-security.md T25/SG-14. Per this review's brief the pin decision itself is settled; I only verified the implementation matches the documented decision, which it does.
- Ordinal-vs-culture comparison is explicitly tested, closing the one way a "consistent" key could quietly stop being consistent.

### Per-DoD-item verdicts (checkable from the diff)

1. Termination mechanism proven - PASS (real SDK reflection plus real-host 404/-32001 test)
2. Keyed on (accountSub, instanceId, pin), credential-sourced - PASS
3. Header-less clients unaffected - PASS (category-scoped plus sink-liveness-guarded test)
4. Positive control (same account and instance replaces) - PASS
5. Both halves provably able to fail (plants) - PASS (13 plants, all RED then restored GREEN; the P2-to-P10 story is itself evidence the plant discipline was taken seriously)
6. No synchronised app release required - PASS (same evidence as item 3)
7. Opaque-token validation (length, charset, CRLF) - PASS
8. Never logged beside bearer - PASS (log-hygiene test proves the replace actually ran before asserting absence, avoiding a vacuous "logged nothing" pass)
9. Exact header name, no X- prefix - PASS
10. Retention lifecycle recorded, fed to w5-6 - PASS per PR body (in-memory only, process-lifetime); I did not independently verify the claimed cross-repo comment on AI-Game-Dev-Server#599, that is outside this repo and diff
11. Delivery chain stated hop-by-hop - PASS per PR body; correctly does NOT claim production yet
12. Version bumps lockstepped - PASS, confirmed 8.1.0 to 8.2.0 in all 5 files (McpPlugin.Common.csproj, Consts.cs, McpPlugin.Server.csproj, server.json x2, McpPlugin.csproj)
13. Rollback recorded - PASS per PR body, textual only
14. G-SERVER-MCP respected - PASS per PR body, textual
15. Build and tests pass - PASS, CI green on all 5 checks; the automated Test Results comment independently confirms +61 tests and 1688 total, matching the PR body's own count

### Non-blocking observations (medium-depth reuse/efficiency lens)

1. Minor, already justified in-code, not asking for a change: the project pin is parsed twice per initialize request (once in McpSessionTokenMiddleware.InvokeAsync, again in StreamableHttpTransportLayer's new block). The in-code comment explains why (avoiding a dependency on AsyncLocal flow across PerSessionExecutionContext, the same class of bug that previously broke SelectionToolContext). Cost is a trivial string-split on a short path, only on initialize. Flagging only so it stays visible to a future reader who might simplify it back to the ambient read.
2. Observation, not a defect: TrackedIdentityCount has no explicit upper bound independent of the SDK's own MaxIdleSessionCount / session churn. This rides on resource limits that already exist elsewhere and are out of this task's scope; I looked for a way to grow the reaper's dictionary without also creating a real, already-bounded MCP session and did not find one, so this is not a finding.

---

Depth/gating note for the orchestrator: medium is advisory-only per code-review.md, so nothing above blocks merge under this run's gating regardless of severity, and nothing above reaches blocking-severity in my judgment either. Given this PR's own stated blast radius (merge triggers an automatic, irreversible NuGet publish of 8.2.0 to nuget.org), I would flag to the orchestrator that advisory-only gating at medium is a narrower net than the stakes described in this run's brief - a run-configuration observation, not something I can or should act on from inside the review itself.

PR #223, opened 2026-09-05
T2 chain fixtures: format v1, recorder/canonicaliser/diff, null-engine --replay

IvanMurzak:
Taskflow `2026-09-03-chain-testing`, task **`p2-record-replay-format`** (group `R-mpd`, sequence 2, after `p1-null-engine-host`).

Defines ONCE the T2 record/replay contract (D13): the fixture format engine legs record, the canonicalisation that makes a diff mean *contract change*, and the null-engine host's `--replay` mode that serves a recorded tool set with no engine present. Consumers are `p2-record-leg-{unity,godot,unreal}` (record + diff) and sibling C's `p2-replay-leg` (App/cloud T1 legs run the host in `--replay`).

## What landed

| | |
|---|---|
| `docs/chain-fixtures.md` | Schema 1: **F1** file, **F2** canonicalisation, **F3** masking, **F4** caps, **F5** diff, **F6** replay, plus the battery format and the vendoring rule. |
| `scripts/chain_fixture.py` | ONE stdlib-only Python file (3.9+), the canonical copy engine repos vendor byte-identically to `.github/scripts/chain_fixture.py`. Subcommands `record`, `canonicalize`, `diff`, `hash`. |
| `McpPlugin.NullEngine/src/Replay/` | `FixtureLoader.cs` (the C# half of F2–F4 + the replay lookup), `ReplayTool.cs` (one per fixture tool line), `RawDump.cs` (the in-process recorder). |
| `McpPlugin.NullEngine/src/Program.cs` | Argument dispatch only: `--replay`, `--dump-raw`, `--battery`, `--engine`, `--engine-version`, `--surface`; exit codes 2 (usage) and 4 (fixture invalid). |
| `tests/chain-fixtures/null-engine/` | The format's reference sample — a battery exercising **all 12** tools, and the fixture recorded from it. |
| `McpPlugin.Server.Tests/Chain/`, `McpPlugin.Tests/Chain/` | 24 new tests. |
| `.gitattributes` | New file — see *Why* below. |

**No product code touched**: `McpPlugin/`, `McpPlugin.Common/`, `McpPlugin.Server/`, the csproj pins, `test-pull-request.yml`, `DemoConsoleApp/` and the golden vectors are all untouched. The only csproj change is a `ProjectReference` + one linked source file in `McpPlugin.Tests` (no package version changed). Engine recorders serialise the **existing** `McpPlugin.Common` types, so engine test code keeps compiling against the pinned 8.3.0 DLLs.

## Two findings the round-trip forced out

**1. The transport rewrites numbers.** The SignalR payload serializer converts every JSON **number** to a string on the way to the plugin: a client sending `{"kb":1}` reaches the tool as `{"kb":"1"}`, recursively (`{"payload":{"number":7,"flag":true}}` arrives as `{"payload":{"number":"7","flag":true}}`; booleans and strings keep their kind). The first replay round-trip therefore came back `replay-miss-args` on **every call carrying a number** while the argument-free calls matched. `args_hash` is now taken over the number-as-raw-text form on both sides, so `1` and `"1"` hash the same — exactly the identification the transport already makes. Pinned by `ArgsHashIdentifiesANumberWithItsWireForm` and by plant 8 below.

**2. `errorKind` never reaches an MCP client.** `ExtensionsTool.ToCallToolResult` maps only `IsError` / `Content` / `StructuredContent`, so the kind is observable **in-process only**. That, plus `meta.recorder`, are the two *recorder-scoped* fields, and DoD 3's byte-identity is only achievable once both are neutralised. They are confined to a `--cross-recorder` flag, so the strict diff an engine leg runs against its own fixture stays strict on both.

> **Deviation from the letter of the spec, flagged deliberately.** F3 lists three masked meta fields; `meta.recorder` had to join them (under `--cross-recorder` only) for two recorder kinds to be comparable at all, which DoD 3 requires. Documented in `docs/chain-fixtures.md` §F3.

**Why `.gitattributes`.** This repository has `core.autocrlf` on and, until now, no attributes file. F1 fixes the fixture encoding at UTF-8/LF and DoD 3/4 compare whole files byte-for-byte — so without it a Windows checkout would rewrite the committed fixture to CRLF and every byte comparison would fail on that leg alone. `*.jsonl`, `*.py` and `tests/chain-fixtures/**` are pinned to `eol=lf`; `ReferenceFixture_IsAlreadyCanonical_AndCoversTheWholeCatalogue` asserts the committed fixture carries no CR.

## Evidence

**Full suite, both TFMs, both projects — local, Release:**

```
Passed! - Failed: 0, Passed:  821, Skipped: 0, Total:  821 - McpPlugin.Server.Tests.dll (net9.0)
Passed! - Failed: 0, Passed:  821, Skipped: 0, Total:  821 - McpPlugin.Server.Tests.dll (net8.0)
Passed! - Failed: 0, Passed: 1018, Skipped: 0, Total: 1018 - McpPlugin.Tests.dll (net8.0)
Passed! - Failed: 0, Passed: 1018, Skipped: 0, Total: 1018 - McpPlugin.Tests.dll (net9.0)
```

**Plants — each run, RED output captured, then reverted:**

| # | DoD | Mutation | Result |
|---|---|---|---|
| a | 5(a) | `ping` response `pong` → `PONG` in the committed fixture | `diff` **exit 3**, `contract-change: … differ`, diff hunk names `null-engine/ping` |
| b1 | 5(b) | only `meta.recorded_at` → `2099-01-01T00:00:00Z` | `diff` **exit 0** |
| b2 | 5(b) | only `meta.mcp_plugin_version` → `8.3.0-ws.gdeadbee` | `diff` **exit 0** |
| b3 | 5(b) | only `meta.plugin_version` → `0.0.0-ws.gdeadbee` | `diff` **exit 0** |
| b-ctl | — | *control*: `meta.surface` → `runtime`; `meta.engine_version` → `9999` | `diff` **exit 3** both — so exit 0 above is not "the meta line is ignored" |
| b-ctl2 | — | *control*: fixture vs itself | `diff` **exit 0** |
| c | 5(c) | drop `RE_WINDOWS_PATH` from `chain_fixture.py`'s mask chain | `CanonicalizeParityTests.PythonAndCSharpProduceTheSameCanonicalBytes(mixed)` **RED**, showing `win C:\Users\bob\a.txt` (Python) vs `win <path>` (C#) |
| d | 5(d) | 40 KB response via `battery-oversize.json` | recorded as `{"$hash":"sha256:a5cce41c…","$bytes":82105}`; tampering with the digest ⇒ `diff` **exit 3**; replay answers `replay-oversize: sha256:a5cce41c…` with `isError:true` |
| e | 5(e) | call a tool the fixture publishes but never exercised | `replay-miss: null-engine/echo`, `isError:true` at the MCP client; wrong args ⇒ `replay-miss-args: null-engine/ping sha256:f7e972f7…` |
| f | 5(f) | delete the `ReplayTool.RegisterAll` call in `Program.cs` | `Replay_RoundTripsByteIdentically_AndServesTheFixturesToolCount` **RED** (`registered=0`, tools count 0) |
| g | extra | make `RawDump` record `mimeType` on text blocks (drift from `ExtensionsContentBlock.ToContent`) | `InProcessDumpAndMcpRecord_AgreeAfterCanonicalization` **RED** with the intended message ("RawDump's mirror … has drifted from the server") |
| h | extra | remove `NumbersAsText` from the C# `ArgsHash` | replay round-trip **RED** (`replay-miss` in the re-recording) **and** 5 of 7 `args_hash` parity cases RED |

**Positive halves (DoD 6).** Every "X did not happen" assertion is paired in the same fixture family with a demonstration that the path *can* produce X:

* Masking: `MaskingRemovesVolatileValues_AndTheRawInputProvesTheyWereThere` asserts the **raw** fixture contains an absolute Windows path, an absolute POSIX path, a GUID and an ISO timestamp *before* asserting the canonical form does not — and then asserts what masking must **not** eat (`slept_ms`, `/variable`, a `null` inside a payload, `2.50`/`1e3`/`-0`, astral-plane characters).
* Cross-recorder parity: the **strict** diff is asserted to fail *first*, on exactly `recorder` and `errorKind`, so byte-identity under `--cross-recorder` cannot be explained by the two files being trivially equal.
* F4 cap: the oversize fixture carries an **un-capped** response beside the capped one.
* F4 refusal: a 180 KB fixture canonicalises fine before the 360 KB one is refused.
* Provenance mask: two non-provenance meta fields still redden.
* Replay fidelity: the re-recording is asserted to contain no `replay-miss` before the byte comparison is credited.

**Manual end-to-end** (local `DemoWebApp` on 127.0.0.1:8791 + a real null-engine process):

```
record over /mcp            -> tools=12 calls=14
--dump-raw in-process       -> tools=12 calls=14
canonicalize --for-diff --cross-recorder both -> BYTE-IDENTICAL (11268 bytes each)
diff (strict, no --cross-recorder)            -> exit 3 (recorder + errorKind)  [the control]
--replay <F1> -> record F3  -> diff exit 0, --for-diff forms byte-identical (11268 bytes)
```

## DoD map

1. ✅ `docs/chain-fixtures.md` F1–F6, testably. 2. ✅ `record` over `/mcp`, N tools == the hand-written set (`RecordOverMcp_…`). 3. ✅ `--dump-raw` + `canonicalize` byte-identical to `record` (`InProcessDumpAndMcpRecord_…`) — see the `--cross-recorder` note above. 4. ✅ Replay round-trip byte-identical, consumer `tools/list` == the fixture's count. 5. ✅ Plants (a)–(f) in the table. 6. ✅ Positive halves above. 7. ✅ `dotnet test` green on all four legs locally; python-gated tests skip **by name** (`Skip = "python3 not on PATH"`) and `ChainToolchainAvailabilityTests` / `PythonGatedTests_ExistAndAreNotSkippedOnLinux` fail on Linux if any of them skipped, and assert count ≥ 1 everywhere. 8. ✅ This body + the CI run below.

https://claude.ai/code/session_01JY6ujdyE1PSsHX4toD1aWG


IvanMurzak:
## CI evidence (DoD 7 / 8)

**Run [33935843528](https://github.com/IvanMurzak/MCP-Plugin-dotnet/actions/runs/33935843528)** — head `5cab9ce`, which is this branch's final **code** (everything after it is documentation only). All four jobs `success`:

| job | result |
|---|---|
| `test (ubuntu-latest)` | ✅ success |
| `test (windows-latest)` | ✅ success |
| `test (macos-latest)` | ✅ success |
| `golden-vector-parity` | ✅ success |

**The python-gated tests actually ran, they did not skip.** From the ubuntu leg of the first run ([33934387090](https://github.com/IvanMurzak/MCP-Plugin-dotnet/actions/runs/33934387090), same test set):

```
Passed com.IvanMurzak.McpPlugin.Server.Tests.Chain.ChainToolchainAvailabilityTests.PythonGatedTests_ExistAndAreNotSkippedOnLinux
Passed com.IvanMurzak.McpPlugin.Tests.Chain.CanonicalizeParityTests.PythonGatedTests_ExistAndAreNotSkippedOnLinux
Passed com.IvanMurzak.McpPlugin.Tests.Chain.CanonicalizeParityTests.PythonAndCSharpProduceTheSameCanonicalBytes(name: "mixed", …)
Passed com.IvanMurzak.McpPlugin.Tests.Chain.CanonicalizeParityTests.ArgsHashIdentifiesANumberWithItsWireForm
…
```

Counted from each leg's log — **48 `Chain` tests passed, 0 skipped, on every OS** (48 = 24 tests × 2 TFMs):

| leg | `Passed …Chain…` | `Skipped …Chain…` |
|---|---|---|
| ubuntu-latest | 48 | 0 |
| windows-latest | 48 | 0 |
| macos-latest | 48 | 0 |

That "0 skipped" is not incidental — `ChainToolchainAvailabilityTests.PythonGatedTests_ExistAndAreNotSkippedOnLinux` is a plain `[Fact]` (a `[PythonFact]` would skip for the very condition it exists to detect) that **fails on Linux** if any python-gated test skipped, and asserts the census is non-empty everywhere so the "none skipped" claim cannot be satisfied by there being none.

## Two follow-up commits after the first green

**`39cc038` — the two fixture readers must ACCEPT the same files.** Found reviewing my own diff, not by a failing test: `FixtureLoader.Load` reads with `File.ReadAllText(path, Encoding.UTF8)`, whose `StreamReader` strips a UTF-8 BOM by itself, while `chain_fixture.py` decoded plain `utf-8` and left it in the first line. A BOM'd fixture therefore **replayed happily and was refused by the tool that gates CI** — the wrong way round. Both sides now strip it, and the C# side strips it in `ReadLines` so `Canonicalize(text)` and `Load(path)` accept the same inputs. Pinned by `ALeadingBomIsAcceptedAndIgnoredByBothImplementations`, positive half first (the file really does start with `EF BB BF`); planted by reverting the Python side to `utf-8` ⇒ RED with `chain_fixture.py must accept a BOM'd fixture. exit=4`.

**`5cab9ce`, `cb50c04` — documentation only.** The tolerated-on-input set, and the seam an engine's own in-process recorder implements (the three `ToolRouter.Call` branches, the per-block-kind field set, the `ResponseErrorKind` vocabulary, and that a raw dump is F1-*shaped* rather than canonical) — `p2-record-leg-{unity,godot,unreal}` cannot run `McpPlugin.NullEngine` inside an editor, so those rules had to be readable from the spec rather than only from `RawDump.cs`.

## Also verified by hand

* `chain_fixture.py hash --file` (the vendoring check `p2-record-leg-*` will run) agrees digit-for-digit with an independent `hashlib.sha256` of the same bytes, gives a **different** digest for a different file, and exits **4** on a missing one.
* No engine process is leaked: the `Terminate orphan process: … (dotnet)` lines in the job logs appear identically on the **pre-null-engine** baseline run (`task/a5-skill-meta-gate-semantics`, 2026-09-01) and in `golden-vector-parity`, which spawns no engine at all — they are MSBuild/VSTest nodes.

https://claude.ai/code/session_01JY6ujdyE1PSsHX4toD1aWG

